PR #1, opened 2017-10-25
Testing info endpoints

rhiGriff:
Tests are running in a pipeline on Jenkins, can be run locally if all services are running locally.


config.py
```diff
@@ -0,0 +1,64 @@
...
+
+    RAS_BACKSTAGE_INFO = RAS_BACKSTAGE_SERVICE + '/info'
+
+    RAS_PARTY_SERVICE_HOST = os.getenv('RAS_PARTY_SERVICE_HOST', 'localhost')
+    RAS_PARTY_SERVICE_PORT = os.getenv('RAS_PARTY_SERVICE_PORT', 8081)
+    RAS_PARTY_SERVICE = '{}://{}:{}'.format(RAS_PROTOCOL,
+                                            RAS_PARTY_SERVICE_HOST,
+                                            RAS_PARTY_SERVICE_PORT)
+
+    RAS_PARTY_INFO = RAS_PARTY_SERVICE + '/info'
```
rjweeks: I think these could be reduced to one constant "INFO_ENDPOINT"

views/features/get_info_endpoints.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Get info endpoints
+
+  Scenario Outline: User hits service get info endpoint
+    Given the user requests <service> endpoint info
```
rjweeks: Given the system requests info endpoint

views/features/steps/get_info_endpoints.py
```diff
@@ -0,0 +1,90 @@
+import nose
+import requests
+
+from config import Config
+from behave import given, when
+from flask import json
+
+
+@given('the user requests secure message endpoint info')
```
rjweeks: Reduce @given to one for all services

Procfile
```diff
@@ -0,0 +1 @@
+web: python run.py
```
benjefferies: Do we need the Procfile if we aren't deploying to CF?

runtime.txt
```diff
@@ -0,0 +1 @@
+python-3.6.1
```
benjefferies: Do we need runtime.txt if we are not pushing to CF?

benjefferies:
Would it be more intuitive to developers if the tests were under a directory called `tests` instead of `views`

views/features/steps/get_info_endpoints.py
```diff
@@ -0,0 +1,90 @@
...
+def step_impl_requests_sm_endpoint_info(context):
+    context.response = requests.get(Config.RAS_SECURE_MESSAGE_SERVICE + Config.RAS_INFO)
+
+
+@given('the user requests frontstage endpoint info')
+def step_impl_requests_fs_endpoint_info(context):
```
benjefferies: Would it be more readable if the method name reflected the step?

views/features/steps/common.py
```diff
@@ -0,0 +1,7 @@
+import nose
+from behave import then
+
+
+@then('a success status code (200) is returned')
+def step_impl_success_returned(context):
```
benjefferies: `step_impl_success_returned` -> `success_returned`. Also it would be good to rename the file to something so it's more cohesive and less likely to be a dumping ground for things with no home. Maybe response_status.py?

PR #2, opened 2017-11-02
Basic ras-frontstage-api test

liamtoozer:
- Added basic check for ras-frontstage-api.
- Removed ras-api-gateway tests.

### DIFF
--- a/config.py
+++ b/config.py
@@ -41,8 +41,3 @@ class Config(object):
                                              RAS_DJANGO_SERVICE_HOST,
                                              RAS_DJANGO_SERVICE_PORT)
 
-    RAS_API_GATEWAY_SERVICE_HOST = os.getenv('RAS_API_GATEWAY_SERVICE_HOST', 'localhost')
-    RAS_API_GATEWAY_SERVICE_PORT = os.getenv('RAS_API_GATEWAY_SERVICE_PORT', 8083)
-    RAS_API_GATEWAY_SERVICE = '{}://{}:{}'.format(RAS_PROTOCOL,
-                                                  RAS_API_GATEWAY_SERVICE_HOST,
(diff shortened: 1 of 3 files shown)


gemmai95:
I am assuming the readme will be updated at some point?

config.py
```diff
@@ -41,8 +41,8 @@ class Config(object):
...
                                              RAS_DJANGO_SERVICE_PORT)
 
-    RAS_API_GATEWAY_SERVICE_HOST = os.getenv('RAS_API_GATEWAY_SERVICE_HOST', 'localhost')
-    RAS_API_GATEWAY_SERVICE_PORT = os.getenv('RAS_API_GATEWAY_SERVICE_PORT', 8083)
-    RAS_API_GATEWAY_SERVICE = '{}://{}:{}'.format(RAS_PROTOCOL,
-                                                  RAS_API_GATEWAY_SERVICE_HOST,
-                                                  RAS_API_GATEWAY_SERVICE_PORT)
+    RAS_FRONTSTAGE_API_SERVICE_HOST = os.getenv('RAS_FRONTSTAGE_API_SERVICE_HOST', 'localhost')
+    RAS_FRONTSTAGE_API_SERVICE_PORT = os.getenv('RAS_FRONTSTAGE_API_SERVICE_PORT', 8083)
+    RAS_FRONTSTAGE_API_SERVICE = '{}://{}:{}'.format(RAS_PROTOCOL,
```
benjefferies: We should probably just have one environmental variable for the apis. Something like RAS_FRONTSTAGE_API = os.getenv('RAS_FRONTSTAGE_API', 'http://localhost:8083'). I think it's unnecessary to have each part of the url split up

liamtoozer: 👍 This would help to cut back on deployment config overrides as well. I'll add to the TODO list.

PR #3, opened 2017-12-12
Adding all info endpoints

rhiGriff:


Pipfile
```diff
@@ -0,0 +1,26 @@
...
+verify_ssl = true
+name = "pypi"
+
+
+[packages]
+
+
+[dev-packages]
+
+Flask = "*"
```
andrewmil: These should be bound to specific versions of packages

system_tests/features/steps/get_info_endpoints.py
```diff
@@ -0,0 +1,200 @@
+import nose
+import requests
+
+from config import Config
+from behave import given, when
+from flask import json
+
+
+@given('the system requests action endpoint info')
```
benjefferies: The BDD/gherkin syntax suggests that the given step should set up the state https://martinfowler.com/bliki/GivenWhenThen.html, https://github.com/cucumber/cucumber/wiki/Given-When-Then.

This would mean the given step would look more like
```python
@given('the action service is running') # This is probably an empty step because the service is already running
@when('I request the action info') # This would make the http request to the endpoint
@then('I retrieve the action service info') # This would check the info is as expected
```

Pipfile
benjefferies: We shouldn't need any of the flask dependencies
```
Flask = "*"
Flask-RESTful = "*"
Flask-SQLAlchemy = "*"
Flask-Cors = "*"
```

Pipfile
```diff
@@ -0,0 +1,26 @@
...
+Flask-RESTful = "*"
+Flask-SQLAlchemy = "*"
+Flask-Cors = "*"
+behave = "*"
+nose = "*"
```
benjefferies: Do we need `nose` as a package?

system_tests/features/steps/get_info_endpoints.py
```diff
@@ -0,0 +1,200 @@
...
+    response = json.loads(context.response.text)
+    nose.tools.assert_equal(response['name'], 'casesvc')
+    nose.tools.assert_equal(response['version'], '10.49.1-SNAPSHOT')
+
+
+@when('the collection exercise endpoint info is returned')
+def collection_exercise_endpoint_info_is_returned(context):
+    response = json.loads(context.response.text)
+    nose.tools.assert_equal(response['name'], 'collectionexercisesvc')
+    nose.tools.assert_equal(response['version'], '10.49.2-SNAPSHOT')
```
benjefferies: We don't really want to be asserting the version of the service. In the CD world every commit will be a new version of the app and would break the tests which is not what we want. I think checking for a 200 status code would be good enough

system_tests/features/get_info_endpoints.feature
```diff
@@ -1,16 +1,25 @@
 Feature: Get info endpoints
 
   Scenario Outline: System hits service get info endpoint
-    Given the system requests <service> endpoint info
-    When  the <service> endpoint info is returned
+    Given the <service> service is running
```
VibhorMehra: Suggest we change the `Given` statement to `Given services are running`. This makes it common to all examples. If we want to separate them, then we might need to use step matchers e.g. https://jenisys.github.io/behave.example/step_matcher/using_matchers.html

system_tests/features/steps/get_info_endpoints.py
```diff
@@ -0,0 +1,89 @@
+import requests
+
+from config import Config
+from behave import given, when
+
+
+@given('the ${service} service is running')
```
VibhorMehra: Suggest changing this to 
`@given('services are running')`

system_tests/features/steps/get_info_endpoints.py
```diff
@@ -0,0 +1,89 @@
...
+def service_running():
```
VibhorMehra: Suggest changing this to
`def service_running(context):`

system_tests/features/steps/get_info_endpoints.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+@when('the system requests party endpoint info')
+def requests_party_endpoint_info(context):
+    context.response = requests.get(Config.PARTY_SERVICE + Config.INFO)
+
+
+@when('the system requests response operations ui endpoint info')
+def requests_response_operations_ui_endpoint_info(context):
+    context.response = requests.get(Config.RESPONSE_OPERATIONS_UI + Config.INFO)
```
VibhorMehra: Response Operations UI does not have an `info` endpoint. However changing this to the following should return status code 200
`context.response = requests.get(Config.RESPONSE_OPERATIONS_UI)`

liamtoozer: I think it might be ras-backstage-ui which doesn't have an `info` end point currently? I was able to hit the info end point for response-operations-ui in the int space. Could we use the above suggestion for the ras-backstage-ui instead?

system_tests/features/steps/get_info_endpoints.py
benjefferies: If the behave framework is designed not to pass context when it's unused it might be best to not define the parameter. It's generally not a good idea to have unused parameters. If we have to have a parameter I think the general pythonic way is to underscore it? `def service_running(_):`

system_tests/features/steps/get_info_endpoints.py
VibhorMehra: ahh.. I was thinking this is the old ROPs, which we us for RM integration tests. Agree that we don't need this change as the new ROPs has an `info` endpoint. 
We can include `ras-backstage-ui` (if needed), not sure if it will be replaced by the new ROPs as well.

system_tests/features/steps/get_info_endpoints.py
VibhorMehra: Agreed using `_` instead of `context` in this case.

system_tests/features/get_info_endpoints.feature
```diff
@@ -1,16 +1,25 @@
 Feature: Get info endpoints
 
   Scenario Outline: System hits service get info endpoint
-    Given the system requests <service> endpoint info
-    When  the <service> endpoint info is returned
+    Given services are running
+    When  the system requests <service> endpoint info
     Then   a success status code (200) is returned
```
benjefferies: Looks like the then step is missing in info_endpoint.py

rhiGriff: The then step is in the common.py

PR #4, opened 2017-12-13
Add BDD test for viewing surveys

benjefferies:
Added test for https://trello.com/c/4uPzDISJ/23-us001b-view-survey-data-8pts

To test
1. Checkout response-operations-ui
1. Hack `surveys.py` to look like the example below
1. Install and run response-operations-ui with `pipenv install && pipenv run python run.py`
1. Run tests with be `pipenv run behave`

  ```
@app.route('/', methods=['GET'])
def view_surveys():
    survey_list = [
  {
    "id": "cb0711c3-0ac8-41d3-ae0e-567e5ea1ef87",
    "shortName": "RSI",
    "longName": "Monthly Business Survey - Retail Sales Index",
    "surveyRef": "023",
    "legal_basis": "Statistics of Trade Act 1947"
  },
  {
    "id": "cb0711c3-0ac8-41d3-ae0e-567e5ea1ef88",
    "shortName": "BRUS",
    "longName": "Business Register and Umployment Survey",
    "surveyRef": "222",
    "legal_basis": "Statistics of Trade Act 1947"
  }
]
    return render_template('surveys.html', survey_list=survey_list)
  ```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,20 +9,74 @@ pipeline {
         }
     }
 
+
+    tools {nodejs "Node 6.10.3"}
+
     stages {
         stage('clone') {
             steps {
-                git url: "https://github.com/ONSdigital/ras-integration-tests.git", branch: "testing-info-endpoints"
+                git url: "https://github.com/ONSdigital/ras-integration-tests.git", branch: "${env.BRANCH_NAME}"
             }
         }
         stage('pip install') {
+
             steps {
-                sh 'pip install -r requirements.txt'
+                sh 'pip install pipenv'
+                sh 'pipenv install --dev'
+                sh 'npm install -g phantomjs-prebuilt'
             }
         }
-        stage('test') {
+        stage('system tests') {
+            environment {
+                RESPONSE_OPERATIONS_UI_SUFFIX = credentials('CF_DOMAIN_SUFFIX')
+                RESPONSE_OPERATIONS_UI_HOST = "response-operations-ui-python-int${RESPONSE_OPERATIONS_UI_SUFFIX}"
+                RESPONSE_OPERATIONS_UI_PORT = "80"
+                SECURE_MESSAGE_SERVICE_HOST = "ras-secure-messaging-int${RESPONSE_OPERATIONS_UI_SUFFIX}"
+                SECURE_MESSAGE_SERVICE_PORT = "80"
+                FRONTSTAGE_SERVICE_HOST = "ras-frontstage-int${RESPONSE_OPERATIONS_UI_SUFFIX}"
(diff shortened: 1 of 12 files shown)


behave.ini
```diff
@@ -1,4 +1,4 @@
 [behave]
-paths=system_tests/features
+paths=acceptance_tests/features
```
benjefferies: We need to look at how we are going to run tests separately

Jenkinsfile
```diff
@@ -12,17 +12,22 @@ pipeline {
...
         stage('clone') {
             steps {
-                git url: "https://github.com/ONSdigital/ras-integration-tests.git", branch: "testing-info-endpoints"
+                git url: "https://github.com/ONSdigital/ras-integration-tests.git", branch: "${env.BRANCH_NAME}"
             }
         }
         stage('pip install') {
             steps {
-                sh 'pip install -r requirements.txt'
+                sh 'pipenv install'
```
jcox-dev: Does this require the --dev flag?

andrewmil: Since everything in this repo is related to testing should we have anything in dev-packages? We could just have it all in packages

benjefferies: It could be dev packages or normal packages to be honest it probably doesn't matter much because it's not going to be deployed anywhere. Happy to put in dev packages if you think it makes sense

acceptance_tests/pages/survey.py
```diff
@@ -0,0 +1,20 @@
+from acceptance_tests import browser
+from config import Config
+
+
+def go_to():
+    browser.visit(Config.RESPONSE_OPERATIONS_UI)
```
benjefferies: 👍 has this been updated in config.py too?

rhiGriff: Yes it was a duplicate so I removed the duplicate

Jenkinsfile
```diff
@@ -9,20 +9,34 @@ pipeline {
         }
     }
 
+    tools {nodejs "Node 6.10.3"}
+
+    environment {
+        RESPONSE_OPERATIONS_UI_URL = 'http://response-operations-ui-python-ci.apps.devtest.onsclofo.uk/'
```
andrewmil: Why do we set this here when the rest of the urls are configured using environment variables?

benjefferies:
I think this is in a place it can be merged back unless there is anything else to do? Thoughts @rhiGriff @liamtoozer ?

The acceptance_tests are currently failing due to the set up of the surveys not being implemented but the system_tests which check the info endpoints are available are passing. Check the jobs out here http://jenkins.rmdev.onsdigital.uk:8080/view/RAS%20Int/job/Test%20Pipeline/job/23-us001b-view-survey-data/

PR #5, opened 2018-01-03
Us002 load and view collection exercises

liamtoozer:
PR blurb courtesy of #4

Added test for https://trello.com/c/TZoUzqkn/24-us002b-view-collection-exercises-13pts

To test
1. Checkout response-operations-ui
2. Hack `surveys.py` to look like the example below (note: this will mean that scenario `Ensure collection exercise exists for a survey` will always pass due to this)
3. Install and run response-operations-ui with `pipenv install && pipenv run python run.py`
4. Run tests with be pipenv run behave
```
@app.route('/surveys/<short_name>', methods=['GET'])
def view_survey(short_name):
    survey_details = {
  "collection_exercises": [
        {
          "id": "14fb3e68-4dca-46db-bf49-04b84e07e77c",
          "name": "QBS_1803",
          "scheduledExecutionDateTime": "2017-09-10T23:00:00.000+0000",
          "exerciseRef": "1803",
          "userDescription": "9 March 2018"
        },
        {
          "id": "14fb3e68-4dca-46db-bf49-04b84e07e77c",
          "name": "QBS_1806",
          "scheduledExecutionDateTime": "2017-09-10T23:00:00.000+0000",
          "exerciseRef": "1806",
          "userDescription": "15 June 2018"
        },
        {
          "id": "14fb3e68-4dca-46db-bf49-04b84e07e77c",
          "name": "QBS_1809",
          "scheduledExecutionDateTime": "2017-09-10T23:00:00.000+0000",
          "exerciseRef": "1809",
          "userDescription": "14 September 2018"
        },
        {
          "id": "14fb3e68-4dca-46db-bf49-04b84e07e77c",
          "name": "QBS_1812",
          "scheduledExecutionDateTime": "2017-09-10T23:00:00.000+0000",
          "exerciseRef": "1812",
          "userDescription": "14 December 2018"
        }
  ],
  "survey": {
    "id": "cb0711c3-0ac8-41d3-ae0e-567e5ea1ef87",
    "longName": "Quarterly Business Survey",
    "shortName": "QBS",
    "surveyRef": "139",
    "legalBasis": "Statistics of Trade Act 1947"
  }
}

    return render_template('survey.html',
                           survey=survey_details['survey'],
                           collection_exercises=survey_details['collection_exercises'])
```


config.py
```diff
@@ -60,6 +60,7 @@ class Config(object):
     RESPONSE_OPERATIONS_UI_HOST = os.getenv('RESPONSE_OPERATIONS_UI_HOST', 'localhost')
     RESPONSE_OPERATIONS_UI_PORT = os.getenv('RESPONSE_OPERATIONS_UI_PORT', 8085)
     RESPONSE_OPERATIONS_UI = '{}://{}:{}'.format(PROTOCOL, RESPONSE_OPERATIONS_UI_HOST, RESPONSE_OPERATIONS_UI_PORT)
+    RESPONSE_OPERATIONS_UI_COLLECTION_EXERCISE = '{}/surveys/{}'.format(RESPONSE_OPERATIONS_UI, '{}')
```
benjefferies: We should probably just put this in `collection_exercise.py` as it's not really configuration as you can't override it and then we only need to format it once

acceptance_tests/features/pages/collection_exercise.py
```diff
@@ -0,0 +1,34 @@
+from acceptance_tests import browser
+from config import Config
+
+
+def go_to(survey):
+    browser.visit(Config.RESPONSE_OPERATIONS_UI_COLLECTION_EXERCISE.format(survey))
```
benjefferies: Based on the comment below it would look something like this f'{RESPONSE_OPERATIONS_UI}/surveys/{survey}'

acceptance_tests/features/steps/view_collection_exercise_data.py
```diff
@@ -0,0 +1,37 @@
...
+def the_internal_user_can_view_all_collection_exercises_for_a_survey(_):
+    # Validate collection exercise table headers
+    table_headers = collection_exercise.get_table_headers()
+    required_headers = ['Period', 'Shown to respondent as']
+    assert table_headers == ' '.join(required_headers)
+
+    # Validate the collection exercises
+    collection_exercises = collection_exercise.get_collection_exercises()
+    for ce in collection_exercises:
+        assert ce['exercise_ref'].value
```
benjefferies: Might be worth having an assertion that checks the values are assigned to the right columns?

acceptance_tests/features/pages/collection_exercise.py
rhiGriff: We have an option of using either .format or f'{} it makes no difference as to which format we use

acceptance_tests/features/steps/view_collection_exercise_data.py
benjefferies: The reasons I think we should be asserting values here is because:
1. A user acceptance testing this change would be checking for the exercise_ref and user_description to be in the correct columns
2. If there are no collection exercises displayed I believe this test will still pass which is obviously wrong.

Summing this up we should probably have an assertion for the expected number of collection exercises or at least that len(collection_exercises) > 0 and collection_exercises[0]['exercise_ref'] == X and collection_exercises[0]['user_description'] == Y

rhiGriff: @benjefferies We checked with Harinder regarding this this morning and have been told to check that there is a value there only not a specific value

benjefferies: Spoken to Harinder and because there isn't a test that already covers this it would be something we should check for

acceptance_tests/features/view_collection_exercise.feature
```diff
@@ -0,0 +1,43 @@
+Feature: View Collection Exercise
```
andrewmil: I think we could change the names of these feature files. 
This test is actually for a view of a single survey and the previous one was for viewing the list of surveys
The next story is around a single view of a collection exercise

acceptance_tests/features/steps/view_collection_exercise_data.py
```diff
@@ -0,0 +1,49 @@
+from behave import given, when, then
+
+from acceptance_tests.features.pages import collection_exercise
+
+
+@given('collection exercises for {survey} exist in the system')
```
benjefferies: It might be nicer to read if the given, when, thens are grouped by the scenario

acceptance_tests/features/view_collection_exercises.feature
```diff
@@ -0,0 +1,43 @@
+Feature: View Collection Exercise
```
andrewmil: Exercises?

acceptance_tests/features/pages/collection_exercise.py
```diff
@@ -0,0 +1,34 @@
+from acceptance_tests import browser
```
andrewmil: Should this file be called collection_exercise**s**?

acceptance_tests/features/steps/view_collection_exercises_data.py
```diff
@@ -0,0 +1,49 @@
+from behave import given, when, then
+
+from acceptance_tests.features.pages import collection_exercise
+
+
+@given('collection exercises for {survey} exist in the system')
```
benjefferies: It might be nicer to read if the given, when, thens are grouped by the scenario

liamtoozer:
[US002b: View Collection Exercise(s) (13pts)](https://trello.com/c/TZoUzqkn/24-us002b-view-collection-exercises-13pts)

PR #7, opened 2018-01-10
Us005 view events

andrewmil:
This was branched off of [us003-view-collection-exercise-details](https://github.com/ONSdigital/ras-integration-tests/tree/us003-view-collection-exercise-details)

Added an integration test to check if an internal user can view events for the BRES 201712 collection exercise.

The dates used in testing were taken from a slightly out of date version of the collection exercise service and the correct first reminder time wasn't known so they will have to be updated before merging.

This PR depends on changes in [response-operations-ui](https://github.com/ONSdigital/response-operations-ui/pull/16) and [ras-backstage](https://github.com/ONSdigital/ras-backstage/pull/30) as well as changes being made in the collection exercise service for [US004](https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/168132632/US004+Load+Events+L)

[Trello](https://trello.com/c/ZtQUGyXs/34-us005-view-events-25pts-l-5-days)
[Confluence](https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/160235597/US005+View+Events+L)


andrewmil:
[US005: View Events (25pts = L = 5 days)](https://trello.com/c/ZtQUGyXs/34-us005-view-events-25pts-l-5-days)

acceptance_tests/features/steps/view_collection_exercise_details.py
```diff
@@ -4,10 +4,15 @@
...
 
 @given('the 1803 collection exercise for the QBS survey has been created')
-def bres_2017_exists(_):
+def qbs_1803_exists(_):
     pass
 
 
+@given('the 1803 collection exercise events for the QBS survey has been created')
+def bres_2017_events_exist(_):
+    pass
```
benjefferies: how are the events being loaded? if i run these tests on my mac the test will fail?

andrewmil: I was following the previous pattern we'd used in these tests so you have to run the tests with the assumption that surveys/collection_exercises/events are already in the system.

Would probably be good to have a local version that did the set up for this on local machines or did them through posting to endpoints during the tests

rhiGriff: This would depend on the story itself if it's the same as the collection exercise data being loaded then the tests would be the same, and the data being loaded would be an environmental setup not a test setup.

benjefferies: Have we solved the way the application is going to load the data? I am concerned that without that being done tests are going to fail

rhiGriff: Vibhor and I are working on a way of loading data for another test once we have figured this out we will communicate on how data will be loaded.

acceptance_tests/features/steps/view_collection_exercise_details.py
```diff
@@ -16,6 +21,17 @@ def internal_user_views_2017_bres_collection_exercise(_):
...
+    assert ce_details['survey_info'] == "139 QBS"
+    assert ce_details['survey_title'] == "Quarterly Business Survey"
     assert ce_details['period'] == "1803"
     assert ce_details['user_description'] == "9 March 2018"
+
+
+@then('the user is able to view the event dates for that collection exercise')
+def internal_user_can_view_bres_2017_collection_exercise_events(_):
+    ce_events = collection_exercise_details.get_collection_exercise_events()
+    assert ce_events['mps'] == "Mon 11 Sep 2017 23:00 GMT"
```
liamtoozer: I guess the data for this will be seeded from here: https://github.com/ONSdigital/rm-tools/blob/3dd720da3f4a21d339f7e9e5284f3dcfb04c4aa0/collex-loader/data/139_2018_schedule.csv#L2 
Would this need to be updated before merging? Not certain what's happening with the data sorry.

andrewmil: This is a good point, I'm not sure what the agreement is on where we get our test data from/how we can be consistent using it

PR #11, opened 2018-01-16
Us008-10 preview load and view sample

tej99:
Acceptance tests for stories US008, US009 and US0010.
Preview, upload and view sample.
Waiting for:
https://github.com/ONSdigital/rm-sample-service/pull/17
https://github.com/ONSdigital/response-operations-ui/pull/26


tej99:
[US010: View sample in context of a CE](https://trello.com/c/9zdn4R3Q/26-us010-view-sample-in-context-of-a-ce)

acceptance_tests/features/load_sample.feature
```diff
@@ -0,0 +1,19 @@
...
+
+  Scenario: Cancel preview sample
+    Given the user is presented with sample details for QBS 1803
+    When the user cancels the sample
+    Then the sample details are reset
+
+  Scenario: Load sample
+    Given the 201803 collection exercise for the QIFDI survey has been created
+    When the internal user navigates to the collection exercise details page for QIFDI 201803
+    Then the user is able to load the sample
```
rhiGriff: No new line at end of file

andrewmil:
@benjefferies I'm pretty sure this happens because you can only have one sample loaded at a time and you can't upload another one once there's one in the database

common/request_handler.py
```diff
@@ -0,0 +1,26 @@
...
+
+import requests
+from requests.exceptions import ConnectionError
+from structlog import wrap_logger
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def request_handler(method, url, auth=None, headers=None, json=None, data=None, files=None, fail=True):
```
benjefferies: Seems overkill to add this in when we only need to `requests.post(url, auth=auth, files=files)`

tej99: files, fail and json are the only ones not being used atm. I've removed fail but kept files and json since there will be use cases for them in the future.

common/request_handler.py
```diff
@@ -0,0 +1,26 @@
...
+        if method == 'GET':
+            response = requests.get(url, auth=auth, headers=headers)
+        elif method == 'POST':
+            response = requests.post(url, auth=auth, headers=headers, json=json, data=data, files=files)
+        elif method == 'PUT':
+            response = requests.put(url, auth=auth, headers=headers, json=json)
+        else:
+            logger.error('Invalid request method', method=str(method), url=url)
+    except ConnectionError as e:
+        logger.error('Failed to connect to external service', method=method, url=url, exception=str(e))
```
rhiGriff: This should be a logger.exception, you then won't need the as e and won't be needed as a parameter

common/request_handler.py
tej99: Oh i think i misunderstood, i agree probably no need for a request handler... i'll remove it.

andrewmil: The only reason we've had request handlers in the other repos is so we can wrap `RequestException` errors into one function. Otherwise we need a try/except every time we make a request somewhere. Not sure what the best way to go here is though

PR #18, opened 2018-01-26
Add pyenv pipenv docs

benjefferies:
Add pyenv and pipenv as prerequisites for clarification


benjefferies:
@andrewmil it might be helpful to some if it was. I've also made a couple more changes

PR #23, opened 2018-01-31
Usxxx internal sign in

rhiGriff:
This PR is to get us334 in. The wording and completion of the sign in scenarios will be on a different PR. PLEASE DO NOT delete the branch after merging!! 😃


acceptance_tests/features/sign_in_internal.feature
```diff
@@ -0,0 +1,17 @@
+Feature: Internal user signs in
+  As an internal user
+  I need to sign in to the SDC system
+  So that I can access my SDC account and carry out a number of actions
+
+  Background: Internal user is already signed in
```
andrewmil: Do we need the background here? The user is signing in in the tests so doesn't need to sign in beforehand I'd think

rhiGriff:
@andrewmil there will be tests for sign out etc in another PR this one is only a quick fix and will be tidied up in the other PR

PR #25, opened 2018-02-01
Hot-fix Internal sign in and out

rhiGriff:
This is a quick fix for the sign in and sign out for us334. The tests will be cleaned up and any formatting etc will be dealt with in another PR.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -25,4 +25,3 @@ python_version = "3.6"
 
 [packages]
 
-sqlalchemy = "*"
--- a/Pipfile
+++ b/Pipfile
@@ -21,3 +21,8 @@ sqlalchemy = "==1.2.2"
 [requires]
 
 python_version = "3.6"
+
+
+[packages]
+
+sqlalchemy = "*"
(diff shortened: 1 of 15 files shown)


Pipfile
```diff
@@ -21,3 +21,8 @@ sqlalchemy = "==1.2.2"
 [requires]
 
 python_version = "3.6"
+
+
+[packages]
+
+sqlalchemy = "*"
```
benjefferies: We shouldn't need sqlalchemy

rhiGriff: Wasn't meant to push it

acceptance_tests/features/load_sample.feature
```diff
@@ -3,17 +3,23 @@ Feature: Load sample
    I need to be able to load the sample
    So that a sample file exists in the system
 
+  Background: Internal user is already signed in
```
benjefferies: This should be a before_step https://pythonhosted.org/behave/api.html#environment-file-functions

PR #26, opened 2018-02-06
Improve initial data setup

andrewmil:
Updated initialisation scripts and the makefile so that now only one command has to be run before running the acceptance tests, `make setup`

This command first pulls `rm_tools/collex-loader` and adds the required collection exercises and events
It then runs the `set_up_ce_execute.sql` script which sets up the 201801 collection exercise for execution
It then adds a collection instrument to the 201801 Bricks CE, previously this was done manually

After this command the acceptance tests can be run.
The next step after this would be to generalise setting up this collection exercise so it's easier for us to add them for future tests


set_up_ce_execution.py
```diff
@@ -1,4 +1,19 @@
...
-    database_controller.reset_rm_database('resources/database/set_up_ce_execute.sql')
+    logging.info("Adding initial data required for executing collection exercise")
+    # Add data required to execute collection exercise
+    database_controller.execute_rm_sql('resources/database/set_up_ce_execute.sql')
+    # Retrieve collection exercise for BRICKS 2018 and upload collection instrument for it
+    bricks_2018_ce = collection_exercise_controller.get_collection_exercise('cb8accda-6118-4d3b-85a3-149e28960c54',
+                                                                            '201801')
+    ci_path = 'resources/collection_instrument_files/064_0001_201803.xlsx'
+    collection_instrument_controller.upload_collection_instrument(bricks_2018_ce['id'], ci_path)
+    logging.info(f"Collection exercise {bricks_2018_ce['id']} can be executed")
```
rhiGriff: Logging needs to be updated to conform with the SDC standards

PR #27, opened 2018-02-06
Generalise setting up collection exercise

andrewmil:
This PR should make it possible to prepare and execute multiple collection exercises before running the acceptance tests, the steps taken were:

- Generalise functions used when preparing/executing collection exercise
- Add sql template for adding future collection exercises
- Execute bricks 201812 collection exercise before tests (this will be used in US060 integration tests)


controllers/sample_controller.py
```diff
@@ -4,7 +4,7 @@
 from config import Config
 
 
-def load_sample(test_file):
-    browser.visit('{}/surveys/{}/{}'.format(Config.RESPONSE_OPERATIONS_UI, 'Bricks', '201801'))
+def load_sample(survey_name, period, test_file):
+    browser.visit(f'{Config.RESPONSE_OPERATIONS_UI}/surveys/{survey_name}/{period}')
```
rhiGriff: These were being kept consistent across the repo, I think it's worth keeping it consistent.

acceptance_tests/features/environment.py
```diff
@@ -16,27 +24,32 @@ def before_all(context):
...
-    sample_controller.load_sample(test_file=test_file)
-    collection_exercise_controller.execute_collection_exercise()
+    sample_controller.load_sample('bricks', '201801', test_file)
+    sample_controller.load_sample('bricks', '201812', test_file)
+    collection_exercise_controller.execute_collection_exercise(survey_id='cb8accda-6118-4d3b-85a3-149e28960c54',
+                                                               period='201801')
+    collection_exercise_controller.execute_collection_exercise(survey_id='cb8accda-6118-4d3b-85a3-149e28960c54',
+                                                               period='201812')
+    logger.info('Waiting for collection exercises execution process to finish...')
+    time.sleep(240)
```
benjefferies: Hmm we really, really want to avoid sleeps in the code. Is there and API we can poll at the least to check for collection exercises

andrewmil: I agree, there's a large (and inconsistent) period of time between the execution of the collection exercise and the generation of iac codes.
We could maybe make a call to some service to check if these have been created, I'll look into it

andrewmil: After some conversations with @benjefferies it looks like there isn't an easy fix to this currently. I'm adding an issue on the repo + a tech debt card. It definitely isn't something that should be left for long

PR #28, opened 2018-02-07
BDD us011 display state of a CE

emilioward:
Added the integration tests for user story 11 displaying the state of a collection exercise

These tests wont pass currently if it is merged they are dependant on the backend stories passing data to respondent-ops ui

https://trello.com/c/kJO9QkLd

### DIFF
--- a/acceptance_tests/features/view_collection_exercises_data.feature
+++ b/acceptance_tests/features/view_collection_exercises_data.feature
@@ -13,11 +13,17 @@ Feature: View Collection Exercise
       | survey_id | survey_title              | survey_abbreviation | survey_legal_basis           |
       | 139       | Quarterly Business Survey | QBS                 | Statistics of Trade Act 1947 |
     And the internal user can view all collection exercises for QBS
-      | period | shown_to_respondent_as |
-      | 1803   | 9 March 2018           |
-      | 1806   | 15 June 2018           |
-      | 1809   | 14 September 2018      |
-      | 1812   | 14 December 2018       |
+      | period | shown_to_respondent_as | status  |
+      | 1803   | 9 March 2018           | Created |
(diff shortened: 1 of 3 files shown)


acceptance_tests/features/view_collection_exercises_data.feature
```diff
@@ -13,11 +13,17 @@ Feature: View Collection Exercise
       | survey_id | survey_title              | survey_abbreviation | survey_legal_basis           |
       | 139       | Quarterly Business Survey | QBS                 | Statistics of Trade Act 1947 |
     And the internal user can view all collection exercises for QBS
-      | period | shown_to_respondent_as |
-      | 1803   | 9 March 2018           |
-      | 1806   | 15 June 2018           |
-      | 1809   | 14 September 2018      |
-      | 1812   | 14 December 2018       |
+      | period | shown_to_respondent_as | status  |
```
benjefferies: I think we should have a separate scenario for the status

emilioward:
[US011 Display State of a Collection Exercise [M]](https://trello.com/c/kJO9QkLd/29-us011-display-state-of-a-collection-exercise-m)

acceptance_tests/features/view_collection_exercises_data.feature
emilioward: I did have that originally but I discussed it with @rhiGriff and Harinder. They mentioned to add the scenario onto the existing scenario for viewing survey attributes, as its all part of the same page but to add an `And` for the viewing of states.

benjefferies: I'd disagree with that justification, scenario testing isn't per page, it's per scenario or feature. If one feature fails only one scenario should fail this way it clear which features of the application are working or have regressed

PR #32, opened 2018-02-09
Integration tests for us043

tej99:
Feature is in master branch
As an: 
internal User
I need: 
to be able to 'find' a Reporting Unit (RU Ref for Business Surveys)
So that: 
I can easily locate the RU to deal with any issues


tej99:
[US043: INT Find an RU [Large]](https://trello.com/c/WyRWFKXq/38-us043-int-find-an-ru-large)

acceptance_tests/features/find_reporting_unit.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Find Reporting Unit
+   As an internal user
+   I need to be able to find a Reporting Unit
+   So that I can easily locate the RU to deal with issues
```
andrewmil: I think we should be consistent with saying 'RU' or 'Reporting unit' In general I'm leaning towards the latter

acceptance_tests/features/pages/reporting_units.py
```diff
@@ -0,0 +1,16 @@
+from acceptance_tests import browser
```
andrewmil: I think this file should be called `reporting_units_search` if we're going to have a file in this folder per page

acceptance_tests/features/steps/find_reporting_unit.py
```diff
@@ -0,0 +1,24 @@
...
+    reporting_units.search_reporting_unit('49900000005')
+
+
+@then('the Reporting Unit is displayed back to the user')
+def internal_user_presented_with_correct_ru(_):
+    rus = reporting_units.get_reporting_units()
+    assert '49900000005' in rus and 'OFFICE FOR NATIONAL STATISTICS' in rus
+
+
+@when('the internal user enters a RU Name')
```
andrewmil: Maybe have this when above the `then` statements?

acceptance_tests/features/pages/reporting_units.py
tej99: It's the /reporting-units page which is why I named it reporting_units.py the reporting unit details page should be called reporting_units_details.py to match the current pattern we have with collection_exercise.py and collection_exercise_details.py.

PR #34, opened 2018-02-12
Delete cookie to start new session between tests

benjefferies:
* Deleting cookies after scenario rather than step will allow downstream tests not to be affected


benjefferies:
The problem with that is if the sign out feature/functionality breaks all scenarios/features will start to fail. The only reason we currency log out is not to test features/scenarios but so that the next scenario can log in. We can log the user out via the UI but I don't see the benefit of logging the user out via the UI for every scenario.

I am not saying that we shouldn't have a scenario/test exercising the logout functionality just that it's not needed in every scenario/test

acceptance_tests/features/steps/authentication.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@when('they click the sign out link')
+@then('the internal user signs out')
+def signed_out_internal(_):
+    browser.find_by_id('sign-out-btn').click()
+
+
+@given('The internal user is already signed in')
+def signed_in_rops(_):
```
rhiGriff: I don't think we should be using rops here, I think it should be internal

acceptance_tests/features/steps/authentication.py
```diff
@@ -0,0 +1,32 @@
+from behave import given, then, when
+
+from acceptance_tests.features.pages import sign_in_respondent
+from acceptance_tests.features.pages import sign_in_internal
+from acceptance_tests import browser
+from config import Config
+
+
+@given('the respondent is signed into their account')
+def signed_in_frontstage(_):
```
rhiGriff: I don't think we should be using frontstage here, I think it should be respondent

acceptance_tests/features/pages/sign_out_internal.py
```diff
@@ -3,3 +3,6 @@
 
 def signed_out_successfully_message():
     browser.find_by_id('successfully-signed-out')
+
+def sign_out():
+    browser.find_by_id('sign-out-btn').click()
```
rhiGriff: No new line at end of file

acceptance_tests/features/steps/authentication.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@given('The internal user is already signed in')
+def signed_in_rops(_):
+    sign_in_internal.go_to()
+    # Only attempt to sign in if not already signed in otherwise implicitly redirected to homepage
+    if '/sign-in' in browser.url:
+        sign_in_internal.enter_correct_username()
+        sign_in_internal.enter_correct_password()
+        sign_in_internal.click_internal_sign_in_button()
```
rhiGriff: We should also have a sign out function for the internal users too

benjefferies: See https://github.com/ONSdigital/ras-integration-tests/pull/34/files#diff-da1a9cfc9613bab3228b43cdf6b14473R8

rhiGriff: Frontstage doesn't have a sign-out-btn id for signing out so this shouldn't be working.

PR #40, opened 2018-02-15
Add a 'make test' to makefile

insacuri:


Makefile
```diff
@@ -22,3 +21,5 @@ system_tests:
 
 acceptance_tests: system_tests
 	pipenv run behave acceptance_tests/features # This will only run the acceptance tests
+
+test: system_tests acceptance_tests
```
benjefferies: Isn't this the same as acceptance_tests, are we just creating another alias?

andrewmil: yeah this does exactly the same thing it seems like?

insacuri: Fair point, does anything else actually use make acceptance_tests for anything or is it only devs that are using it?  I could do ```test: acceptance_tests``` and that would be more correct.  Would that be better?

benjefferies: Not 100% sure, it's worth checking in jenkins for the integration tests

insacuri: And it does do the same thing, but ```make tests``` is really easy to type whereas ```make acceptance_tests``` is a pain to type.  It's personal preference but I think it'll make everyones lives immeasurably better!

PR #42, opened 2018-02-16
Enable stdout logging and add more logging

benjefferies:


acceptance_tests/__init__.py
```diff
@@ -1,7 +1,12 @@
+import logging
 import os
 
 from splinter import Browser
+from structlog import configure
+from structlog.stdlib import LoggerFactory
 
+logging.basicConfig()
+configure(logger_factory=LoggerFactory())
```
rhiGriff: Why do we need to do this?

benjefferies: It's not actually logging anything at the moment. This will fix that

rhiGriff: Do we need this can't we use one of the --no-capture commands for running the tests? https://pythonhosted.org/behave/behave.html#cmdoption--no-capture

benjefferies: Didn't work unless I am missing something?

rhiGriff: Not sure, I have heard it works

PR #44, opened 2018-02-16
us018 load variable event dates

emilioward:
Tests for the load variable event dates user story 

Please test this against the changes made to RM-tools and not using `make setup`

Run load.py and load_events.py aginst the changes made in rm-tools, then run `pipenv run python set_up_ce_execution.py` and `make acceptance tests` to test against the new data.


emilioward:
[US018: INT Load variable event dates for a collection exercise [M]](https://trello.com/c/n4uTwf2f/43-us018-int-load-variable-event-dates-for-a-collection-exercise-m)

PR #46, opened 2018-02-19
Run all tests in travis

benjefferies:


.travis.yml
```diff
@@ -7,9 +12,12 @@ cache:
 install:
   - pip install pipenv==8.3.2
   - pipenv install --dev --deploy
+  - npm install -g phantomjs-prebuilt
+  - docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD";
 
 script:
   - pipenv check --style . --max-line-length 120
+  - make start_services setup acceptance_tests stop_services
```
jcox-dev: Missing the docker network on Travis: `docker network create rasrmdockerdev_default`

.travis.yml
```diff
@@ -7,9 +12,13 @@ cache:
 install:
   - pip install pipenv==8.3.2
   - pipenv install --dev --deploy
+  - npm install -g phantomjs-prebuilt
+  - docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD";
+  - docker network create rasrmdockerdev_default
```
jcox-dev: Will this cause the install step to fail if the network already exists (exit code 1)?

benjefferies: I don't think it should ever exist in travis because it should spin us up a new instance every time. Part of the reason for putting it in .travis.yml instead of the make file

acceptance_tests/features/pages/reporting_unit.py
```diff
@@ -31,7 +31,7 @@ def get_associated_collection_exercises():
                 "exercise_ref": row.find_by_name('tbl-ce-period').value,
                 "company_name": row.find_by_name('tbl-ce-company-name').value,
                 "company_region": row.find_by_name('tbl-ce-company-region').value,
-                "status": row.find_by_name('tbl-ce-status').value
+                "status": browser.find_by_name('tbl-ce-status').value
```
jcox-dev: This will need changing back to `row.find_by_name...` I expect

wait_until_services_up.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def retry_if_http_error(exception):
+    print(f'error has occurred: {str(exception)}')
+    return isinstance(exception, RequestException) or isinstance(exception, HealthCheckException)
+
+
+@retry(retry_on_exception=retry_if_http_error, wait_fixed=10000, stop_max_delay=600000, wrap_exception=True)
+def check_services():
+    for port in [Config.ACTION_SERVICE_PORT, Config.BACKSTAGE_SERVICE_PORT,
```
jcox-dev: Using the ports makes sense locally but will this be suitable for testing deployment? (Might make more sense to list the full service config variable for each instead)

acceptance_tests/__init__.py
```diff
@@ -9,6 +9,8 @@
 configure(logger_factory=LoggerFactory())
 
 if os.getenv('HEADLESS', 'True') == 'True':
+    browser = Browser('chrome', headless=True)
+elif os.getenv('HEADLESS', 'phantomjs') == 'phantomjs':
```
JamesGardiner: Do we want to add support for a deprecated method of testing?

acceptance_tests/features/steps/view_reporting_unit_details.py
```diff
@@ -1,7 +1,14 @@
+from logging import getLogger
+
+import time
```
JamesGardiner: time is a stdlib module so should be in first code block

Makefile
```diff
@@ -1,13 +1,27 @@
...
 .PHONY: test system_tests acceptance_tests
 
 install:
 	pipenv install --dev
 
+start_services:
+	git clone --depth 1 ${RAS_RM_REPO_URL} tmp_ras_rm_docker_dev
+	cd tmp_ras_rm_docker_dev\
+	&& make up\
+	&& docker stop oauth2-service\
```
jcox-dev: These next two lines (oauth2-service) can be removed with the changes to the docker Makefile now in master.

Pipfile
```diff
@@ -16,6 +16,7 @@ chromedriver-installer = "*"
 testfixtures = "*"
 sqlalchemy = "==1.2.2"
 "psycopg2" = "==2.7.3.2"
+retrying = "*"
```
jcox-dev: We should pin this (despite the low maintenance of the project).

wait_until_services_up.py
JamesGardiner: You could do:

```Python
def check_status(port):
    try:
        resp = requests.get(f'http://localhost:{port}/info')
        resp.raise_for_status()
    except Exception:
        raise HealthCheckException(port)

[check_status(k) for k, v in dict(vars(Config)).items() if '_PORT' in k]
```

Pipfile
benjefferies: Do we need to pin any of these, that's what the pipfile.lock does right?

acceptance_tests/__init__.py
benjefferies: Not adding just keep it for backwards compatibility in jenkins until we sort out the jenkins build

wait_until_services_up.py
benjefferies: Ah, nice. Can't do this now as it points to the whole url unless I change the variable names but this is pretty nice

acceptance_tests/__init__.py
insacuri: Can the jenkins build not handle chrome?  I thought splinter brought drivers for all the browsers along for the ride?

benjefferies:
I've ran this 3 times now and it's passed. Really hoping it's in a stable place and can be merge

acceptance_tests/__init__.py
benjefferies: I don't think splinter pulls the drivers down, we've manually had to pull to driver down in the pipfile. Even if it did I don't think the driver actually equates to the actual browser, we'd need to install chrome on jenkins which isn't impossible just a breaking change that I don't want to introduce in this PR

PR #49, opened 2018-02-21
US028 Set collection exercise ready for live

jcox-dev:
Test that big ol' button.

- ~depends on https://github.com/ONSdigital/response-operations-ui/pull/72~
- ~missing scenarios involving validation (005, 009, 010)~ – separate PR.


acceptance_tests/features/environment.py
```diff
@@ -39,6 +41,29 @@ def before_scenario(_, scenario):
         return
 
 
+def prepare_collection_exercises():
+    # for us028
```
benjefferies: If it's for us028 I wonder if it would be easier to follow if it's set up in the feature given step, what do you think?

Makefile
```diff
@@ -7,12 +7,22 @@ install:
 	pipenv install --dev
 
 start_services:
-	git clone --depth 1 ${RAS_RM_REPO_URL} tmp_ras_rm_docker_dev
+	if [ -d tmp_ras_rm_docker_dev ]; then \
```
benjefferies: 👍

acceptance_tests/features/environment.py
```diff
@@ -8,8 +8,8 @@
 from acceptance_tests.features.pages import sign_out_internal
 from acceptance_tests.features.steps import authentication
 from config import Config
-from controllers import collection_exercise_controller, database_controller, sample_controller
-from controllers import party_controller, django_oauth_controller, case_controller
+from controllers import (collection_exercise_controller, database_controller, sample_controller,
```
andrewmil: Just a quick note here about how we do imports where they end up exceeding line length. This is my preferred method because it's closed to what we do elsewhere in our code. There are some other places where we use / at the end of a line. Either way is good but might be nice to keep things consistent

acceptance_tests/features/environment.py
```diff
@@ -40,6 +41,17 @@ def before_scenario(_, scenario):
         return
 
 
+def before_step(context, step):
+    try:
+        context.survey, context.survey_period = next(tag.split('_')[1:] for tag in context.tags if tag.startswith('ce'))
```
benjefferies: Would the correct way be to parameterise the collection exercise? https://jenisys.github.io/behave.example/tutorials/tutorial03.html

jcox-dev: That would be much more preferable. I just wasn't sure about inserting additional (context) data into the defined acceptance criteria.

resources/database/populate_action_rules.sql
```diff
@@ -0,0 +1,9 @@
+/* Populate action rules for ready for live */
+
```
tej99: Possibly rename this file to be more specific to a survey? rsi_populate_action_rules.sql?

jcox-dev: @tej99 Ah, are they specific to RSI? (It was copy-pasted from elsewhere by @madeye-matt - possibly prod).

tej99: Looks like they are, the casetypedefault is specific to RSI (75b19ea0-69a4-4c58-8d7f-4458c8f43f5c) which is linked to the actionplan which is linked to the action rules.

jcox-dev: @tej99 I'll get it renamed.

PR #50, opened 2018-02-21
Sm101 inbox

insacuri:


controllers/database_controller.py
```diff
@@ -54,6 +56,34 @@ def reset_ras_database():
...
+    with open('resources/database/database_reset_secure_message.sql', 'r') as sqlScriptFile:
+        sql = sqlScriptFile.read().replace('\n', '')
+
+    connection.execute(sql)
+    trans.commit()
+    connection.close()
+
+
+def create_secure_messages():
+    logger.info("Putting party database into known state")
```
benjefferies: Is there a way we can do this via an api?

gemmai95: I agree you should be hitting the SM API for this rather than the database directly I think.

benjefferies: Also I guess the full UI is there to be able to do this so you could drive it off the UI unless there is a reason not to

acceptance_tests/features/inbox_internal.feature
```diff
@@ -0,0 +1,28 @@
+Feature: Internal inbox
+  As an internal user
+  I need to view all inbox messages
+  So that I can work on the messages as required
+
+  Background: Internal user is already signed in
+    Given the internal user is already signed in
+
+  Scenario: If there are no messages the user will be informed of this.
```
gemmai95: Can you add tags to these scenarios? So sm111_s01 etc?

controllers/messages_controller.py
```diff
@@ -0,0 +1,39 @@
...
+        'msg_from': "BRES",
+        'msg_to': ["f62dfda8-73b0-4e0e-97cf-1b06327a6712"],
+        'subject': subject,
+        'body': body,
+        'thread_id': "",
+        'collection_case': "",
+        'survey': "cb0711c3-0ac8-41d3-ae0e-567e5ea1ef87",
+        'ru_id': "c614e64e-d981-4eba-b016-d9822f09a4fb"}
+
+    url = Config.SECURE_MESSAGE_SERVICE + '/message/send'
```
gemmai95: Can we target version 2, instead of one as that only handles BRES and all the new functionality needs to work with version 2?

acceptance_tests/features/pages/inbox.py
```diff
@@ -0,0 +1,54 @@
...
+            'business_name': row.find_by_name('tbl-messages-business').value,
+            'subject': row.find_by_name('tbl-messages-subject').value,
+            'from': row.find_by_name('tbl-messages-from').value,
+            'to': row.find_by_name('tbl-messages-to').value,
+            'received': row.find_by_name('tbl-messages-received').value
+        })
+    return messages
+
+
+def are_headings_correct():
```
benjefferies: Maybe get_headings and asserting the headings in the test will make the test more readable

acceptance_tests/features/pages/inbox.py
```diff
@@ -0,0 +1,54 @@
...
+    table = browser.find_by_id('tbl-messages')
+    headings = table.find_by_tag('thead').find_by_tag('tr')
+    return headings[0].value == "RU_Ref Business name Subject From To Received"
+
+
+def are_messages_in_reverse_chronological_order():
```
benjefferies: Assertions (checking the order of the messages) should be done ideally in the test making the test more readable and also leaving the responsibility of the page object to just access data on the page

acceptance_tests/features/inbox_internal.feature
```diff
@@ -0,0 +1,28 @@
...
+    Given the user has access to secure messaging
+    When they navigate to the inbox messages
+    Then they are informed that there are no messages
+
+  Scenario: User is able to view all Inbox messages.
+    Given the user has access to secure messaging
+    And the secure message database is populated with messages
```
benjefferies: I think the BDD scenarios should be written in a language the user knows. The user wouldn't be talking about messages being populated in the database I guess. More user friendly language might be something like

```
Given the respondent has got messages in their inbox (Implied access to secure messaging)
When they navigate to the inbox messages
Then they are able to view all received messages 
```

controllers/messages_controller.py
insacuri: As talked about, this needs to be a future piece of work as the current inbox also only uses the v1 code.

acceptance_tests/features/inbox_internal.feature
insacuri: Added them

acceptance_tests/features/pages/inbox.py
insacuri: by 'in the test' do you mean inside the steps file?

PR #58, opened 2018-02-26
Check if directory exists before cloning

insacuri:


insacuri:
Can you just raise that as an issue?  I did this with a quick 10 mins of free time that I had and I'm not sure when I'll get another bit of free time to sort it out

jcox-dev:
@insacuri No worries. I already had both implemented in https://github.com/ONSdigital/rasrm-acceptance-tests/pull/49.

insacuri:
@jcox-dev how long is your PR going to take?  This is ready now whereas yours looks like it might take a little while and this is an issue that's a bit of a pain when setting up

PR #59, opened 2018-02-27
Re-order make test to check services before setup

jcox-dev:
Local changes -> bring up services -> check services -> load initial state -> run tests -> bring down services.

Currently: bring up services -> load initial state -> fall over if local style issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ style_tests:
 system_tests:
 	pipenv run behave system_tests/features # This will only run the system tests
 
-acceptance_tests: system_tests
+acceptance_tests:
 	pipenv run behave acceptance_tests/features # This will run the system and acceptance tests
 
-test: start_services setup style_tests acceptance_tests stop_services
+test: style_tests start_services system_tests setup acceptance_tests stop_services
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -39,7 +39,7 @@ style_tests:
 system_tests:
 	pipenv run behave system_tests/features # This will only run the system tests
 
-acceptance_tests: system_tests
+acceptance_tests:
 	pipenv run behave acceptance_tests/features # This will run the system and acceptance tests
```
insacuri: Need to update the comment and the readme if this is no longer going to run the system tests first

Makefile
```diff
@@ -39,7 +39,7 @@ style_tests:
 system_tests:
 	pipenv run behave system_tests/features # This will only run the system tests
 
-acceptance_tests: system_tests
```
tej99: I found running the system_tests before running the acceptance_tests very useful, any reason for getting rid of it?

jcox-dev: Not getting rid of it, just separating them from the acceptance tests as it surely makes more sense to run them before setup and after the health check has been performed.

PR #61, opened 2018-02-28
tests for US062 - access eq survey

tej99:
These tests should run with no errors against master branches.


madeye-matt:
Looks ok to me

PR #72, opened 2018-03-08
Add acceptance tests for reopened case group

benjefferies:
As a: internal user

I need: to change the response status for an RU for a specific CE for a
specific survey to Re-opened

So that: the status reflects the action that I have taken internally and
externally


benjefferies:
Not needed for release 17

PR #76, opened 2018-03-14
Sm124 get thread

atrozzo:
ready for review. Due the missing setup in the acceptance test to obtain uua id and use v2 secure messages API, I could not test a full conversation with an extarnal message. I have tested teh scenarios using just messages sent from internal and V1 API.


acceptance_tests/features/view_full_thread.feature
```diff
@@ -0,0 +1,33 @@
+Feature: View conversation thread
+  As an internal user
+  I need to see all of the previous messages in a conversation
+  So that I have the full context for my reply
+
+  Background: Internal user is already signed in
```
gemmai95: Not sure this step is necessary anymore as the login should be disable for the acceptance tests.

PR #79, opened 2018-03-16
SM139 Filter messages by survey

dcdarrell9:
Added the ability to view messages by filter

[This PR needs to be tested against this PR](https://github.com/ONSdigital/response-operations-ui/pull/92)

### DIFF
--- a/acceptance_tests/features/pages/inbox_internal.py
+++ b/acceptance_tests/features/pages/inbox_internal.py
@@ -4,6 +4,10 @@ from acceptance_tests import browser
 
 
 def go_to():
+    browser.visit(f"{Config.RESPONSE_OPERATIONS_UI}/messages/BRES")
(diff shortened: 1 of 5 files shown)


acceptance_tests/features/pages/inbox_internal.py
```diff
@@ -35,3 +39,11 @@ def get_table_heading():
 
 def get_no_messages_text():
     return browser.find_by_text('No new messages')
+
+
+def get_ashe_radio_button_value():
+    return browser.find_by_id('ASHE')
+
```
tej99: `get_radio_button_for_survey(survey):
          return browser.find_by_id(survey)
`
This might be better than having a function for each individual survey. Also function name is slightly misleading as it's not returning the value it's returning an element object.

PR #87, opened 2018-03-21
Acceptance v2 api

atrozzo:
Replaced the internal message creation via direct API call to V1 service with Selenium Browser navigation that use underneath the V2. 
Increased the report unit go_to page time sleep to 1 second because flaky and suddenly failing for no reason.


controllers/messages_controller.py
```diff
@@ -1,48 +1,18 @@
...
-def create_thread():
-    msg_to = [RESPONDENT_DETAILS.get_respondent_id()]
-    ru_id = RESPONDENT_DETAILS.get_ru_id()
-    create_message(msg_to, "test subject", "test body", ru_id)
-    logger.info("Create a message conversation thread")
-
-
-def _get_jwt():
-    return jwt.encode({'user': 'BRES', 'party_id': 'BRES', 'role': 'internal'}, 'testsecret', algorithm='HS256')
+def create_message():
```
dcdarrell9: This looks nicer. Can we maybe split this out into a create message and then send message or maybe rename it to `create_and_send_message` ?

dcdarrell9: I'm not sure actually, does create_message imply that we create and inbox/sent message?

acceptance_tests/features/pages/reporting_unit.py
```diff
@@ -6,7 +6,7 @@
 
 def go_to(ru_ref):
     browser.visit(f"{Config.RESPONSE_OPERATIONS_UI}/reporting-units/{ru_ref}")
-    time.sleep(0.1)
+    time.sleep(1)
```
benjefferies: Should be waiting for an element on the page instead of an arbitrary wait (I know you didn't add it initially) otherwise this will lead to flaky tests

acceptance_tests/features/steps/inbox_internal.py
```diff
@@ -15,12 +14,12 @@ def verify_messages_link_present(_):
...
 def populate_database_with_messages(_):
-    respondent_id = [RESPONDENT_DETAILS.get_respondent_id()]
-    ru_id = RESPONDENT_DETAILS.get_ru_id()
-    messages_controller.create_message(respondent_id, "test subject", "test body", ru_id)
+
+    messages_controller.create_message()
     # 1 sec sleep so that there is a different timestamp on the message
-    time.sleep(1)
-    messages_controller.create_message(respondent_id, "test subject2", "test body2", ru_id)
+    time.sleep(2)
```
benjefferies: Should be waiting for an element on the page instead of an arbitrary wait (I know you didn't add it initially) otherwise this will lead to flaky tests

acceptance_tests/features/pages/reporting_unit.py
atrozzo: @benjefferies  this test is already flaky that s why I have increased the time. I agree that WebDriverWait would be a better solution but we are talking baout existing code. Can we pleas eopen a tech deb card about it, and leave this oustise this code review ?

benjefferies: https://trello.com/c/n9Q7lPaL/85-acceptance-tests-should-wait-for-elements-to-appear-instead-of-arbitrary-waits-sleeps

PR #88, opened 2018-03-22
Add circle ci config

benjefferies:


jcox-dev:
You'll probably need to use the new `wait_for_services` target before the next step can be run.

.circleci/config.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - run:
+          name: Install dependencies
+          command: |
+            curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash
+            export PATH="~/.pyenv/bin:$PATH"
+            eval "$(pyenv init -)"
+            eval "$(pyenv virtualenv-init -)"
+            pyenv install 3.6.0
+            pip install --user pipenv==8.3.2
```
jcox-dev: Will probably want to drop the pinning or use a later version here.

PR #90, opened 2018-03-23
removed time.sleep from code and replaced with browser find with delay.

atrozzo:
I have tried to remove where ever was possible the call ot the time.sleep function and replaced with the standard one of the browser function that include a delay. In same case I needed a retry, so that I have created a function in the commons that is actually recursive. Please review the code and change as you need, I ll be off fo rthe rest of the week. I have created a pull request as well for check Travis result . locally it s all green.


acceptance_tests/features/pages/reporting_unit.py
```diff
@@ -78,7 +75,8 @@ def click_generate_new_code():
 
 
 def click_edit_details(survey_short_name):
-    browser.find_by_id(f'edit-contact-details-btn-{survey_short_name}').click()
+    if browser.is_element_present_by_id(f'edit-contact-details-btn-{survey_short_name}', wait_time=2):
```
benjefferies: How come we needed to check for the element to be present here?

acceptance_tests/features/steps/inbox_internal.py
```diff
@@ -16,9 +16,10 @@ def verify_messages_link_present(_):
 def populate_database_with_messages(_):
 
     messages_controller.create_message("This is the subject of the message", "This is the body of the message")
-    # 1 sec sleep so that there is a different timestamp on the message
-    time.sleep(2)
-    messages_controller.create_message("This is the subject of the message", "This is the body of the message")
+
+    if browser.is_text_present('Message sent.', wait_time=2):
```
benjefferies: 👍

acceptance_tests/features/steps/set_ready_for_live.py
```diff
@@ -35,11 +36,7 @@ def prepare_collection_exercises(_, survey, period):
         # form type hard coded to 0001 for all ces to simplify testing
         collection_instrument_controller.upload_seft_collection_instrument(ce['id'], ci_path, '0001')
 
-    for i in range(5):
-        state = collection_exercise_controller.get_collection_exercise(s_id, period)['state']
-        if state == 'READY_FOR_REVIEW':
-            break
-        time.sleep(1)
+    browser.is_text_present('READY_FOR_REVIEW', wait_time=2)
```
benjefferies: Not sure if this will work because the page won't be updated asynchronously although the data is

acceptance_tests/features/steps/view_collection_exercise_details_ready_for_review_state.py
```diff
@@ -42,17 +40,11 @@ def navigate_to_rsi_details(_):
 @then('the status of the 201811 collection exercise is Ready for Review')
 def rsi_201811_is_ready_for_review(_):
     collection_exercises = collection_exercise.get_collection_exercises()
-    ce_state = next((ce['exercise_ref'] for ce in collection_exercises if ce['exercise_ref'] == '201811'))
     # Status updated async so wait until updated
-    for i in range(5):
-        if collection_exercise.is_ready_for_review(ce_state):
-            break
-        time.sleep(1)
+    if not browser.is_text_present('Ready for review', wait_time=2):
```
benjefferies: Not sure if this will work because the page won't be updated asynchronously although the data is. It is likely it will always wait for 2 seconds rather that up to 2 seconds

acceptance_tests/features/steps/view_collection_exercise_details_ready_for_review_state.py
```diff
@@ -74,12 +66,6 @@ def load_collection_instruments(_):
 
 @then('the status of the collection exercise is Ready for Review')
 def ce_details_state_is_ready_for_review(_):
-    ce_state = collection_exercise_details.get_status()
-    # Status updated async so wait until updated
-    for i in range(5):
-        if collection_exercise.is_ready_for_review(ce_state):
-            break
-        time.sleep(1)
+    if not browser.is_text_present('Ready for review', wait_time=2):
```
benjefferies: Not sure if this will work because the page won't be updated asynchronously although the data is

acceptance_tests/features/steps/view_collection_exercise_live_state.py
```diff
@@ -39,14 +38,10 @@ def go_live_date_hits(_, survey, period):
 @then('the state of a collection exercise is to be changed to Live')
 @then('they are able to see the Live Status for that Collection Exercise')
 def ce_details_state_is_live(_):
-    ce_state = collection_exercise_details.get_status()
-    for i in range(20):
-        if collection_exercise.is_live(ce_state):
-            break
-        time.sleep(1)
+    if not browser.is_text_present('Live', wait_time=2):
```
benjefferies: Not sure if this will work because the page won't be updated asynchronously although the data is. It is likely it will always wait for 2 seconds rather that up to 2 seconds

PR #91, opened 2018-03-26
SM114 Added scenarios for distinguishing unread messages

AdamHawtin:
Added tests for distinguishing unread messages in the internal users message box.

Trello card:
[SM114](https://trello.com/c/hY01M9Lv/63-sm114-int-identify-read-unread-messages-5)

Confluence page:
[SM114](https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/223314187/SM114+Internal+User+-+Identify+if+Messages+are+Read+Unread+in+the+inbox)

**How to test:**
All acceptance tests should pass with up to date master branches of all services.


acceptance_tests/features/inbox_internal.feature
```diff
@@ -36,3 +36,21 @@ Feature: Internal inbox
     Given the user has access to secure messaging
     When they navigate to the select survey page
     Then they are able to view a list of filter options
+
+  @sm114_s01
+  Scenario: User is able to distinguish unread messages in their inbox
+    Given the user has no messages in their inbox
+    And the user has an unread message in their inbox
```
gavinleeedwards: It'd be impossible for the user to have no messages AND an unread message.  Might be better to switch the logic around a bit to something along the lines of:

Given the user is signed in
And have access to secure messaging
And the user has no messages in their inbox
When the user receives a message from an external user
And the user navigates to their inbox
Then the user is able to see 1 unread message in their inbox

Similar for scenario @sm114_s02

acceptance_tests/features/inbox_internal.feature
```diff
@@ -36,3 +36,21 @@ Feature: Internal inbox
...
+    And the internal user is already signed in
+    When they navigate to the inbox messages
+    Then they are able to distinguish that the message is unread
+
+  @sm114_s02
+  Scenario: Messages are no longer distinguished as unread once they have been viewed
+    Given the user has no messages in their inbox
+    And the user has an unread message in their inbox
```
gavinleeedwards: Similar to @sm114_s01

PR #92, opened 2018-03-27
removed time.sleep from code and replaced with browser find with delay.

insacuri:
This PR is the same as https://github.com/ONSdigital/rasrm-acceptance-tests/pull/90 at it's first commit, but with improvements to the method used to remove the sleeps, rather then stripping it out and getting it wrong.


common/browser_utilities.py
```diff
@@ -0,0 +1,10 @@
+def is_text_present_with_retry(browser, retries: int, text: str, delay: int) -> bool:
```
benjefferies: Don't need to pass in browser it's global, also I think it's worth defaulting the retries and the delay

common/browser_utilities.py
```diff
@@ -0,0 +1,10 @@
...
+    for attempt in range(retries):
+        if not browser.is_text_present(text, wait_time=delay):
+            browser.reload()
+            retry = retries - 1
+            is_text_present_with_retry(browser, retry, text, delay)
```
benjefferies: Do we actually need recursion here. Wouldn't the below work?

```
def is_text_present_with_retry(text: str, retries: int = 5, delay: int = 100) -> bool:
    for attempt in range(retries):
        if not browser.is_text_present(text, wait_time=delay):
            browser.reload()
        else:
            return True

    return False
```

common/browser_utilities.py
benjefferies: Also we could take a lambda rather than just checking for text as this could hide bugs if there is the same text multiple times on the page?

common/browser_utilities.py
```diff
@@ -0,0 +1,10 @@
+from acceptance_tests import browser
+
+def is_text_present_with_retry(text, retries=3, delay=1):
```
jcox-dev: Useful! Could be tidied up by dropping the else:

```python
def is_text_present_with_retry(text, retries=3, delay=1):
    for attempt in range(retries):
        if browser.is_text_present(text, wait_time=delay):
            return True
        browser.reload()
    return False
```

acceptance_tests/features/steps/view_collection_exercise_live_state.py
```diff
@@ -39,14 +38,7 @@ def go_live_date_hits(_, survey, period):
...
 def ce_details_state_is_live(_):
-    ce_state = collection_exercise_details.get_status()
-    for i in range(20):
-        if collection_exercise.is_live(ce_state):
-            break
-        time.sleep(1)
-        browser.reload()
-        ce_state = collection_exercise_details.get_status()
-    assert collection_exercise.is_live(ce_state), ce_state
+    assert is_text_present_with_retry('Live', 2)
```
benjefferies: Do we need more retries here as it was up to 20 before suggesting it might be something that's slow to process it

insacuri: 20?  I just used the numbers that were in the initial PR.  No problems in increasing that though

insacuri:
Branch passed locally twice in a row.

PR #94, opened 2018-03-28
Refactor initial test setup

andrewmil:
General refactor of the initial setup of data for the acceptance tests, i.e. what happens during the `make setup` step and what is run before the tests in `environment.py`

This PR aims to:
- Make it easier to add more collection exercises for future tests
- Stop using the UI to add initial test data, if the UI is to fail we want it to be during the tests for that functionality
- Reuse more code when executing collection exercises
- Standardise sql execution (somewhat)

To test:
Run the tests, there's been no changes with how to start/setup the tests


acceptance_tests/features/environment.py
```diff
@@ -47,38 +37,54 @@ def after_step(context, step):
...
+    else:
+        upload_seft_collection_instrument(collection_exercise['id'],
+                                          'resources/collection_instrument_files/064_201803_0001.xlsx',
+                                          form_type='0001')
+
+    sample_summary = sample_controller.upload_sample(collection_exercise['id'],
+                                                     'resources/sample_files/business-survey-sample-date.csv')
+    collection_exercise_controller.link_sample_summary_to_collection_exercise(collection_exercise['id'],
+                                                                              sample_summary['id'])
+    time.sleep(5)
```
benjefferies: Can we change this time.sleep to a poll mechanism? These sort of poll checks should help with the flakiness of the tests

acceptance_tests/features/environment.py
```diff
@@ -47,38 +37,54 @@ def after_step(context, step):
...
-    logger.info('Waiting for collection exercises execution process to finish',
-                survey_id='cb8accda-6118-4d3b-85a3-149e28960c54', period='201812')
     poll_database_for_iac(survey_id='cb8accda-6118-4d3b-85a3-149e28960c54', period='201812')
-    logger.info('Waiting for collection exercises execution process to finish',
-                survey_id='02b9c366-7397-42f7-942a-76dc5876d86d', period='1809')
     poll_database_for_iac(survey_id='02b9c366-7397-42f7-942a-76dc5876d86d', period='1809')
+    logger.info('Collection exercises have finished executing')
+
+
+def execute_collection_exercise(survey_id, period, ci_type='SEFT'):
```
benjefferies: Is this setup pretty stable now? Used to fail sometime hence the `logger.exception('Failed to setup before running tests', html=browser.html)`

controllers/database_controller.py
```diff
@@ -2,71 +2,45 @@
...
 
 
-def reset_secure_message_database():
-    logger.info("Clearing down secure message database")
-    engine = create_engine(Config.SECURE_MESSAGE_DATABASE_URI)
-    connection = engine.connect()
-    trans = connection.begin()
-
-    with open('resources/database/database_reset_secure_message.sql', 'r') as sqlScriptFile:
+def execute_rm_sql(sql_script_file_path):
```
benjefferies: Is it worth running this like you've done above?

acceptance_tests/features/environment.py
andrewmil: This should be more stable now that we're not using the UI the execute the CE's

acceptance_tests/features/environment.py
andrewmil: I'll have a look into it but might not be changed with this PR. There's quite a lot getting changed already

controllers/database_controller.py
andrewmil: So apparently in the cf environments the different databases have different username/passwords required in the URI. That's why we pass in a specific database URI into the execute_ras_sql, they'll be different for all the RAS databases. For the RM databases we get around this problem by using the cf database tool.

I'd like it to be more consistent ideally but I don't think it's easy to do with the current database setup in cloudfoundry

andrewmil: I do have a couple more changes I'll push to this file to make it a bit more generic though before the PR goes through

controllers/database_controller.py
```diff
@@ -10,50 +10,55 @@
 logger = wrap_logger(logging.getLogger(__name__))
 
 
-def execute_ras_sql(sql_script_file_path, database_uri=Config.DATABASE_URI):
+def execute_ras_sql(sql_script_file_path=None, sql_string=None, database_uri=Config.DATABASE_URI):
```
jcox-dev: 👍

acceptance_tests/features/edit_respondent_details.feature
```diff
@@ -43,46 +44,23 @@ Feature: As an internal user
...
-    And the internal user has found the respondents details
+    Given the respondent with email "test_respondent2@test.com" is enrolled
+    And the internal user has found the second respondents details
     When they change the email address
     And they click save
     Then they are able to save the updated email address
+    And they are presented with confirmation that the changes have been saved
 
   @us058-s002
-  Scenario: The user is able to enter up to 254 characters for an email address
```
benjefferies: How comen we're deleting these tests?

andrewmil: @us058-s001, @us058-s002 and @us058-s003 were functionally all doing the same thing, I can add them back in but it's not testing any new functionality.

@us058-s007 was the pretty much identical as well and is now covering in the final `And` step added to @us058-s001

benjefferies: Let's just keep them separate I think. Smaller more specific tests give better feedback and generally live by the rule of only testing one thing. This doesn't come without downsides like speed but they can be mitigated. Our test time isn't actually a concern at the moment as they run in ~ 5 minutes

After reading the stories they are poorly written and should have been refined before now.

I think `Scenario: The user is able to enter up to 254 characters for an email address` is a technical requirement so shouldn't really be an acceptance criteria.

In the scenario `Scenario: The user is able to edit the email address of a respondent`
`When they change the email address` and `Then they are able to save the updated email address` doesn't describe at all what the user is expecting to see it's just rewording the `when`

andrewmil: :+1: definitely agree on them being poorly written I'll try and separate them out where appropriate then

acceptance_tests/features/edit_respondent_details.feature
```diff
@@ -43,46 +44,23 @@ Feature: As an internal user
 
   @us058-s001
   Scenario: The user is able to edit the email address of a respondent
-    Given the respondent with email "test_respondent@test.com" is enrolled
-    And the internal user has found the respondents details
+    Given the respondent with email "test_respondent2@test.com" is enrolled
+    And the internal user has found the second respondents details
```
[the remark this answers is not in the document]
andrewmil: So originally the `test_respondent@test.com` that was being used here was actually the `example@example.com` user and when it's email was changed it became unverified and this was causing failures for me.
There's now two new users created for these tests, one to change other details, and one to change it's email. I should probably reword this though because it doesn't really make sense from a business point of view

gemmai95: Rather than second respondent can we use their email or name as second respondent doesn't really mean anything.

acceptance_tests/features/steps/edit_respondent_details.py
```diff
@@ -48,7 +51,7 @@ def change_contact_number(_):
 
 @when('they change the email address')
 def edit_email_address(_):
-    edit_respondent_details_form.edit_email_address(Config.RESPONDENT_USERNAME)
+    edit_respondent_details_form.edit_email_address('test_respondent3@test.com')
```
gemmai95: If you are going to refactor these tests you might as well add the email address as a parameter.

acceptance_tests/features/edit_respondent_details.feature
andrewmil: I'm going to use your suggestion below and use the email as a parameter :+1:

acceptance_tests/features/edit_respondent_details.feature
```diff
@@ -4,85 +4,55 @@ Feature: As an internal user
...
   @us058-s001
   Scenario: The user is able to edit the email address of a respondent
-    Given the respondent with email "test_respondent@test.com" is enrolled
-    And the internal user has found the respondents details
+    Given the respondent with email "test_respondent2@test.com" is enrolled
+    And the internal user has found the "test_respondent2@test.com" respondents details
     When they change the email address
     And they click save
-    Then they are able to save the updated email address
+    Then they are presented with confirmation that the changes have been saved
```
rhiGriff: I still think this should not be within this test and that @us058-003 should be brought back in

acceptance_tests/features/steps/edit_respondent_details.py
```diff
@@ -46,9 +36,9 @@ def change_contact_number(_):
     edit_respondent_details_form.edit_contact_number("01633 878787")
 
 
-@when('they change the email address')
-def edit_email_address(_):
-    edit_respondent_details_form.edit_email_address(Config.RESPONDENT_USERNAME)
+@when('they change the email address to "{email}')
```
gemmai95: You are missing a close to the double quotes after the closing }. Other than that looks good.

PR #95, opened 2018-03-28
Set user to active in party service when registering

andrewmil:
The example user `example@example.com` was never being set to active in the party service which wouldn't let them sign in

This is now set in database when a respondent is registered


controllers/database_controller.py
```diff
@@ -117,6 +117,7 @@ def enrol_party(respondent_uuid):
     sql_statement_update_enrolment = f"UPDATE partysvc.enrolment SET status = 'ENABLED' WHERE respondent_id = (SELECT id FROM partysvc.respondent WHERE party_uuid = '{respondent_uuid}');"  # NOQA
     sql_get_case_id = f"SELECT case_id FROM partysvc.pending_enrolment WHERE respondent_id = (SELECT id FROM partysvc.respondent WHERE party_uuid = '{respondent_uuid}');"  # NOQA
     sql_delete_pending_enrolment = f"DELETE FROM partysvc.pending_enrolment WHERE respondent_id = (SELECT id FROM partysvc.respondent WHERE party_uuid = '{respondent_uuid}');"  # NOQA
+    sql_set_account_active = f"UPDATE partysvc.respondent SET status = 'ACTIVE' WHERE party_uuid = '{respondent_uuid}';"  # NOQA
```
gemmai95: There is an endpoint in party that can do this rather than going directly to the database

andrewmil: I think a lot of the register respondent stuff can be done with endpoints rather than this SQL. I'll have a look at adding it to [my refactoring PR](https://github.com/ONSdigital/rasrm-acceptance-tests/pull/94)

PR #107, opened 2018-04-16
reset ce details and update headers

tej99:
Fixing acceptance tests for features:
- Internal user is able to edit/amend collection exercise: Fails after the first run because the CE details need to be reset
- View attributes for a survey: Fails because headers have changed in the ui


tej99:
@jwgwalton yes, looking into it as part of another pr

PR #108, opened 2018-04-16
us106-edit-survey-details

emilioward:
Run alongside:

https://github.com/ONSdigital/rm-survey-service/pull/31
https://github.com/ONSdigital/response-operations-ui/pull/124
https://github.com/ONSdigital/ras-backstage/pull/97


emilioward:
[US106: INT User to Edit/Amend Survey (L)](https://trello.com/c/BGNKc7cM/63-us106-int-user-to-edit-amend-survey-l)

acceptance_tests/features/pages/survey.py
```diff
@@ -27,3 +27,7 @@ def get_surveys():
 def click_qbs_survey_link():
     link = browser.find_by_name('survey-link-QBS')
     link.click()
+
+
+def click_edit_survey_details_button():
+    browser.click_link_by_id('edit-survey-details-199')
```
andrewmil: In general I don't think we should be changing the details for existing surveys. We should be creating a new survey for the test then editing that to prevent us contaminating the data for other tests.
We'd need the create survey endpoints to be present for this so might be good to update in the future

emilioward: @andrewmil Yeah I talked about that with Tej this morning currently these 2 tests:

```Failing scenarios:
  acceptance_tests/features/view_collection_exercises_data.feature:42  Ensure collection exercise exists for a survey -- @1.13 
  acceptance_tests/features/view_survey_data.feature:9  View all default surveys```

fail because they are asserting for survey data that has been changed and we only do the reset before the acceptance tests startup. So I can put a temporary fix in to update these values back to what they were before after the tests for updating survey details have passed.

File: acceptance_tests/features/create_message_internal.feature [hunk 2]
```diff
@@ -1,51 +1,51 @@
...
-    When they select send
-    Then they are navigated to the inbox of messages
-
-  @sm101_s07
-  Scenario: On sending the message the user is to receive confirmation that the message has been sent
-    Given the user is on the send message page
-    And the user has sent a secure message
-    When they are navigated to the inbox of messages
-    Then they are presented with confirmation that the message has been sent
+#Feature: Internal user to send message
```
rhiGriff: Do these need to be commented out?

emilioward: These will be uncommented just did that to make testing easier but accidentally added that in the commit

PR #113, opened 2018-04-27
Add survey list dropdown

dcdarrell9:
- Added dropdown list instead of list of surveys.
- Must be run with [this branch in response-operations](https://github.com/ONSdigital/response-operations-ui/pull/138)


MIslam92:
[Add new surveys (MBS, OFAT, GOVERD) to message filtering (1)](https://trello.com/c/h2gmLVXL/168-add-new-surveys-mbs-ofat-goverd-to-message-filtering-1)

controllers/party_controller.py
```diff
@@ -66,13 +66,13 @@ def add_survey(party_id, enrolment_code):
 
 
 def get_party_by_email(email):
-    logger.debug('Retrieving party by email address', email=email)
+    logger.debug('Retrieving party by email address')
```
insacuri: Why is this being removed?

rhiGriff: The email shouldn't be logged out, no personal information should be logged out not even in tests really.

PR #119, opened 2018-05-02
Make waiting for case to update optional when registering a respondent

andrewmil:
It takes some time for the relevant cases to be updated when we register/enrol a respondent. In some cases we need to wait for this to test but in some cases we don't. 
This PR removes this wait when it is not necessary as it can take some time

### DIFF
--- a/acceptance_tests/features/steps/edit_respondent_details.py
+++ b/acceptance_tests/features/steps/edit_respondent_details.py
@@ -7,8 +7,13 @@ from controllers.party_controller import get_party_by_email
 
 
 @given('the respondent with email "{email}" is enrolled')
-def respondent_ail_is_enrolled(_, email):
+def respondent_is_enrolled(_, email):
     create_respondent(email)
+
+
+@given('the respondent with email "{email}" is enrolled and active')
+def respondent_is_enrolled(_, email):
+    create_respondent(email, wait=True)
 
 
 @given('the internal user chooses to change "{email}" account details')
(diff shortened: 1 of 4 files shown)


acceptance_tests/features/steps/edit_respondent_details.py
```diff
@@ -104,8 +109,8 @@ def error_email_already_in_use(_):
     assert reporting_unit.save_email_error()
 
 
-def create_respondent(email):
+def create_respondent(email, wait=False):
     email_in_use = get_party_by_email(email)
     if not email_in_use:
         register_respondent(survey_id='cb8accda-6118-4d3b-85a3-149e28960c54', period='201801',
-                            username=email, ru_ref=49900000001)
+                            username=email, ru_ref=49900000001, wait_for_case=wait)
```
rhiGriff: Shouldn't be logging the email

insacuri: Doesn't really matter for a test suite does it?

rhiGriff: Shouldn't get into the habit of it, these things have been found to be logged out in prod recently we shouldn't get used to seeing this kind of info in logs

andrewmil: These are all test emails, and logging them here is useful as this is a point where things can go wrong and it's helpful to know which respondent has failed to register

rhiGriff: Apologies I thought I saw it say logger here.

PR #122, opened 2018-05-10
US110 Display "Trading As" in To do and History

AdamHawtin:
Adds acceptance tests for US110

To test run this branch of ras-frontstage locally:
[Us110 dispay trading as #328](https://github.com/ONSdigital/ras-frontstage/pull/328)

Confluence page:
[US110: EXT - Display "Trading As" in To do and History for each RU [M]](https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/376963077/US110+EXT+-+Display+Trading+As+in+To+do+and+History+for+each+RU+M)


AdamHawtin:
[US110: EXT Display "trading as" in To do and History [M]](https://trello.com/c/pXNYDmoS/204-us110-ext-display-trading-as-in-to-do-and-history-m)

tej99:
@emilioward works for we when I run that feature on its own but when I run all the acceptance tests it fails.

emilioward:
@tej99 I actually ran all the tests and individually and even debugged through them and they seemed to pass, couldn't understand why they errored in the first place. Must have been due to some race condition possibly

AdamHawtin:
@emilioward @tej99 The issue was the survey was already being added somewhere else in the tests so party returned a 500 when it tried to add it again. I'll investigate how this is happening but for now my solution is just to catch the exception and carry on since it is actually in the required state already.

The other issue was the wait for case to update was just looking at the first case in a list, I've added a version that looks specifically for the correct case(s) to be actionable.

PR #124, opened 2018-05-11
Fix wait for live

andrewmil:
Changes in how we validate collection exercise event dates mean that the previous way we triggered/tested the live state transition of a collection exercise wont work.
We now create a new collection exercise which transitions into a live state over the course of the test

Works with changes in
[ras-frontstage](https://github.com/ONSdigital/ras-frontstage/pull/330)


gemmai95:
@emilioward Merging master in fixes the first one.

andrewmil:
@emilioward These should be fixed now

PR #146, opened 2018-06-29
Change text for no conversations

ajmaddaford:
# Motivation and Context
Tests updated to match functionality being introduced in https://github.com/ONSdigital/ras-frontstage/pull/352 and https://github.com/ONSdigital/response-operations-ui/pull/170

# What has changed
Text when user (internal and external) has no messages/conversations

# How to test?
- Run ras-frontstage:SM145-view-closed-conversation and response-operations-ui:SM142-internal-user-view-closed-conversations
- Run acceptance tests


shahi645:
FYI @ajmaddaford just caught up with master as, the tests wouldn't even get going because of backstage config

PR #151, opened 2018-07-13
Attempt to fix error creating Chrome

benjefferies:
# Motivation and Context
We occasionally see failures starting chrome with the error
`Exception ConnectionResetError: [Errno 104] Connection reset by peer`.
There seems to be an open bug around this on github https://github.com/SeleniumHQ/selenium/issues/5296
and as a work around adding a retry mechanism to hopefully ignore it
until it's fixed. Also updating libraries in the chance that might fix
it.

# What has changed
Add retry decorator to the method to create Chrome and updated libs

# How to test?
Ensure no regression in acceptance tests

# Links
https://trello.com/c/8N1h5ArZ/154-investigate-chrome-failing-to-start-intermittently


acceptance_tests/__init__.py
```diff
@@ -9,10 +10,16 @@
...
-    browser = Browser('chrome', headless=True, options=chrome_options)
-else:
-    browser = Browser('chrome')
+
+@retry(retry_on_exception=lambda e: isinstance(e, ConnectionError), wait_fixed=1000, stop_max_attempt_number=30)
+def create_browser():
+    if os.getenv('HEADLESS', 'True') == 'True':
+        chrome_options = webdriver.ChromeOptions()
+        chrome_options.add_argument("--no-sandbox")
+        chrome_options.add_argument("--window-size=1920,1200")
```
damien-lloyd: If chrome is running in headless mode, why are we setting the size of the window?

benjefferies: The app isn't yet responsive so there were tests failing. It might be ok now

PR #158, opened 2018-08-06
Round start time down to nearest minute

benjefferies:
# Motivation and Context
SFTP seems to check for a newer file even though the file it's expecting
is there. This could be because the precision of SFTP time is only to
the minute.

# What has changed
Round start time down to the nearest minute

# How to test?
* Run acceptance tests where the test time is in the same minute as the file is put on the SFTP

# Links
https://trello.com/c/Ii0xwuGt/225-add-acceptance-test-for-generating-notification-file


AdeHarrison:
unable to review as collection exercise ci service returns 400

PR #162, opened 2018-08-13
1st stage of button id naming

AdeHarrison:
# Motivation and Context
Make frontstage button id naming consistent.

# What has changed
changed button ids in rasrm-acceptance-tests & ras-frontstage repos

# How to test?
test with ras-frontstage "inconsistant_html_button_id_naming" branch running locally in PyCharm etc... or from local docker image i.e docker build -t frontstage-ade .
all acceptance tests should pass.

# Links
https://github.com/ONSdigital/ras-frontstage/pull/381

### DIFF
--- a/acceptance_tests/features/pages/add_survey.py
+++ b/acceptance_tests/features/pages/add_survey.py
@@ -7,7 +7,7 @@ def go_to():
 
 
 def get_access_survey_button():
-    sign_in_button = browser.find_by_id('SIGN_IN_BUTTON')
+    sign_in_button = browser.find_by_id('sign_in_button')
     sign_in_button.click()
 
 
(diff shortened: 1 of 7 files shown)


liamtoozer:
Looks good to me 👍 Might be worth changing these 2 ID's as well. I think they're both from frontstage but I might be wrong!

https://github.com/ONSdigital/rasrm-acceptance-tests/blob/cd5be67ab9d77dfd0331fd17584149f1333dcadd/acceptance_tests/features/pages/surveys_history.py#L6

https://github.com/ONSdigital/rasrm-acceptance-tests/blob/cd5be67ab9d77dfd0331fd17584149f1333dcadd/acceptance_tests/features/pages/surveys_history.py#L13

AdeHarrison:
I agree @liamtoozer but had to roll back some initial changes as they caused all sort of problems, only changed buttons with explicitly named ID's ("BUTTON", "BTN" etc...) for now - it's not perfect but most of the way there

PR #168, opened 2018-08-22
Changes to handle drop case iac column

AdeHarrison:
# Motivation and Context
to support changes for generating a new access (IAC) code

# What has changed
iac column has been dropped from case table and only caseiacaudit iac column should be used

# How to test?
Test in conjunction with [rm-case-service PR](https://github.com/ONSdigital/rm-case-service/pull/69)

# Links


controllers/database_controller.py
```diff
@@ -46,36 +46,20 @@ def get_iac_for_collection_exercise(collection_exercise_id, social=False):
     else:
         sample_unit_type = "B"
```
damien-lloyd: Sample unit type is always business? Shouldn't this be parametrised to include social as well?

AdeHarrison:
[SUS004: Generate a new access code (IAC) for an LMS case](https://trello.com/c/WTjuaOv5/190-sus004-generate-a-new-access-code-iac-for-an-lms-case)

controllers/case_controller.py
```diff
@@ -51,6 +51,11 @@ def get_b_case(collection_exercise_id, business_id):
                   if case['sampleUnitType'] == 'B')
     logger.debug('Successfully retrieved B case',
                  collection_exercise_id=collection_exercise_id, business_id=business_id)
+
+    iac = get_case_iac(b_case['id'])
```
liamtoozer: There is an IAC flag used `get_cases_by_casegroup` above - could that be used to get the IAC instead? I'm currently getting an IAC of `null` at the moment however when that call to the case service is made, so I might be wrong!

PR #172, opened 2018-09-03
Remove reset of removed tables

benjefferies:
# Motivation and Context
We no longer use the report table or sequence so removing reference to
them. This was removed in
https://github.com/ONSdigital/rm-actionexporter-service/pull/52

# What has changed
Remove reference to table and sequence

# How to test?
Using https://github.com/ONSdigital/rm-actionexporter-service/pull/52 run all the services then `make wait_for_services setup acceptance_tests`


emilioward:
Still getting errors running this @benjefferies

PR #179, opened 2018-09-17
Fix auto go_live dates set only 1 min in future = flakey acceptance tests

NickGrantONS:
# Motivation and Context
The Collection Exercise service will automatically change the state of a collex on the `go_live` date provided that the state transition is valid, but if it's in `EXECUTION_STARTED` because it's still receiving sample units then the state transition will never happen because there's an error and no retry.

We could change the Collection Exercise service, but when would we ever set a `go_live` date one minute in the future and then attempt to upload a sample? Retries & DLQ in Collection Exercise are desirable, but beyond the scope of fixing these flakey acceptance tests.

# What has changed
Updated all the tests which set a `go_live` date 1 minute in the future to 2 minutes.

# How to test?
It's an intermittent fault which is hard to reproduce locally. Try running some really CPU & memory intensive stuff at the same time as the acceptance tests and you might see it. It causes the pipeline to fail pretty regularly, but it's hard to know if we've fixed it because it's intermittent.

# Links
Trello: https://trello.com/c/DvavjcNS/376-bug-investigate-intermittent-failure-of-collex-to-receive-process-all-sample-units-hanging

### DIFF
--- a/acceptance_tests/features/steps/change_response_status.py
+++ b/acceptance_tests/features/steps/change_response_status.py
@@ -23,7 +23,7 @@ def create_collection_exercise(_, survey, period):
     now = datetime.utcnow()
     dates = {
         "mps": now + timedelta(seconds=5),
-        "go_live": now + timedelta(minutes=1),
+        "go_live": now + timedelta(minutes=2),
         "return_by": now + timedelta(days=10),
         "exercise_end": now + timedelta(days=11),
     }
--- a/common/enrolment_helper.py
+++ b/common/enrolment_helper.py
@@ -47,7 +47,7 @@ def generate_collection_exercise_dates():
 
     dates = {
         'mps': now + timedelta(seconds=5),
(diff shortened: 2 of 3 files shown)


acceptance_tests/features/steps/change_response_status.py
```diff
@@ -23,7 +23,7 @@ def create_collection_exercise(_, survey, period):
     now = datetime.utcnow()
     dates = {
         "mps": now + timedelta(seconds=5),
-        "go_live": now + timedelta(minutes=1),
+        "go_live": now + timedelta(minutes=2),
```
shahi645: @NickGrantONS Are we sure 2 minutes is enough, or is this a sort of trial and error process?

NickGrantONS: trial and error, but every minute we add delays the whole acceptance test suite by 1 minute

PR #183, opened 2018-09-25
Fix SUS006 test, use API calls instead of respondent-home-ui

AdamHawtin:
# Motivation and Context

The SUS006 test worked fine locally but it was hitting respondent-home-ui's front end to test certain things, this service is not currently being deployed to CI so that test was failing in the pipeline. This PR replaces the interactions with RH with API calls to the same effect.

# What has changed

* Deleted pages file which references respondent home ui
* Changed interactions with respondent home ui calls to direct API calls

# How to test?

Run the acceptance tests locally without respondent-home-ui running (mimics CI space), they should still pass.

# Links

https://trello.com/c/4ukDtQX9/216-sus006-disable-uac-when-response-status-is-changed-to-data-to-be-deleted-5


AdamHawtin:
[SUS006: Disable UAC when Response Status is changed to 'Data to be deleted' (5)](https://trello.com/c/4ukDtQX9/216-sus006-disable-uac-when-response-status-is-changed-to-data-to-be-deleted-5)

PR #192, opened 2018-10-04
Make notification file test more robust

tej99:
# Motivation and Context

Currently only on enrolment code was getting checked, now we check
all of the expected enrolment code. 
Also the enrolment codes sometimes get written over multiple notification files, now we make sure we check all files instead of just one.

# What has changed

Now checking all enrolment codes are in the notification file.
Also check all notification files for that survey ref and 
collection exercise instead of just one.

# How to test?

Ran ATs

# Links

### DIFF
--- a/acceptance_tests/features/steps/notification_letter.py
+++ b/acceptance_tests/features/steps/notification_letter.py
@@ -1,5 +1,4 @@
 import logging
-import time
 from datetime import timedelta, datetime
 
 import paramiko
@@ -7,10 +6,11 @@ from behave import given, when, then
 from retrying import retry
 from structlog import wrap_logger
 
-from acceptance_tests.features.environment import poll_database_for_iac
 from config import Config
+from controllers import collection_exercise_controller
 from controllers.collection_exercise_controller import create_and_execute_collection_exercise
-
+from controllers.database_controller import get_all_iacs_for_collection_exercise,\
+    poll_collection_exercise_until_ready_for_live_or_live
 
 logger = wrap_logger(logging.getLogger(__name__))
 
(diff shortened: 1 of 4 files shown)


acceptance_tests/features/steps/notification_letter.py
```diff
@@ -1,16 +1,18 @@
 import logging
 import time
+from copy import copy
```
NickGrantONS: unused import

acceptance_tests/features/steps/notification_letter.py
```diff
@@ -64,24 +66,35 @@ def _get_sftp_client():
...
+                if iac in content:
+                    seen_iacs.append(iac)
+
+    if set(seen_iacs) != set(expected_iacs):
+        file_names = [f.filename for f in files]
+        logger.info('Unable to find all iacs', files_found=file_names, seen_iacs=seen_iacs, expected_iacs=expected_iacs)
         raise FileNotFoundError
 
-    return f'{Config.SFTP_DIR}/{filename}'
+    for file in files:
```
NickGrantONS: This is potentially problematic. We don't know _which_ files the IAC codes were in, and we could end up deleting files which didn't have *any* IAC codes we expected in them, which might hamper test debugging

controllers/database_controller.py
```diff
@@ -61,3 +62,40 @@ def get_iac_for_collection_exercise(collection_exercise_id, social=False):
     for row in result:
         iac = row['iac']
     return iac
+
+
+def get_all_iacs_for_collection_exercise(collection_exercise_id, social=False):
```
NickGrantONS: is `get_iac_for_collection_exercise` now dead unused code?

acceptance_tests/features/steps/notification_letter.py
tej99: Okay, is removing the delete the files here and 'reseting' the sftp server at the start of the ATs better?

controllers/database_controller.py
tej99: Nope used in multiple places

PR #196, opened 2018-10-15
Use post case event endpoint instead of casegroup transition

andrewmil:
# Motivation and Context
We are removing the `/casegroups/transitions/...` endpoint so casegroups now need to be transitioned with a case event

# What has changed
Replace calls to the casegroups transitions endpoints with case event posts

# How to test?
Acceptance tests should pass when ran with associated PR's

# Links
[Trello](https://trello.com/c/xkWvMZFk/420-use-case-event-rather-than-case-group-transition-for-change-response-status)

## Associated PR's
[response-operations-ui](https://github.com/ONSdigital/response-operations-ui/pull/250)
[response-operations-social-ui](https://github.com/ONSdigital/response-operations-social-ui/pull/17)
[rm-case-service](https://github.com/ONSdigital/rm-case-service/pull/93)

### DIFF
--- a/controllers/case_controller.py
+++ b/controllers/case_controller.py
@@ -97,17 +97,6 @@ def get_case_by_party_id(party_id):
     return response.json()
 
 
-def update_case_group_status(collection_exercise_id, ru_ref, case_group_event):
-    logger.debug('Updating status', collection_exercise_id=collection_exercise_id, ru_ref=ru_ref,
-                 case_group_event=case_group_event)
-    url = f'{Config.CASE_SERVICE}/casegroups/transitions/{collection_exercise_id}/{ru_ref}'
-    response = requests.put(url, auth=Config.BASIC_AUTH, json={'event': case_group_event})
-
-    response.raise_for_status()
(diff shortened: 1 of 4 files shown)


controllers/case_controller.py
```diff
@@ -97,17 +97,6 @@ def get_case_by_party_id(party_id):
...
-    url = f'{Config.CASE_SERVICE}/casegroups/transitions/{collection_exercise_id}/{ru_ref}'
-    response = requests.put(url, auth=Config.BASIC_AUTH, json={'event': case_group_event})
-
-    response.raise_for_status()
-    logger.debug('Successfully updated status', collection_exercise_id=collection_exercise_id, ru_ref=ru_ref,
-                 case_group_event=case_group_event)
-
-
 def get_case_iac(case_id):
     url = f'{Config.CASE_SERVICE}/cases/{case_id}'
```
shahi645: This is named wrong?

shahi645: Actually looking at case endpoint in case i can see the IAC is returned. Ignore me

PR #210, opened 2018-11-07
Add target to run ALL tests sequentially, in case of false positives

NickGrantONS:
# Motivation and Context
The acceptance tests are not running reliably in Travis or on our local dev machines. Some devs are seeing very frequent (75%+) failures and others are seeing intermittent (~20%) failures, which are false positives, which is making it very difficult to use the acceptance tests to check for regression bugs.

# What has changed
Added a new option to the Makefile called `acceptance_sequential_tests_all` which will run all the tests (including the parallel ones) sequentially, which should be our preferred test to run when we're validating any code changes, until we can understand why the parallel tests fail so frequently locally.

# How to test?
Run `acceptance_sequential_tests_all` and check all 213 scenarios pass.

# Links
Trello: https://trello.com/c/s1SAIKhT/563-create-option-to-run-all-the-acceptance-tests-in-parallel

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,10 @@ acceptance_parallel_tests:
 	export IGNORE_SEQUENTIAL_DATA_SETUP=True; \
 	pipenv run python run_in_parallel.py --command_line_args="${TEST_ARGS}"
 
+acceptance_sequential_tests_all: setup
+	export IGNORE_SEQUENTIAL_DATA_SETUP=False; \
+	pipenv run python run_in_sequence.py --command_line_args="${TEST_ARGS}" --test_tags="~donotskipany"
+
 rasrm_acceptance_tests: rasrm_acceptance_sequential_tests rasrm_acceptance_parallel_tests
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -48,6 +48,10 @@ acceptance_parallel_tests:
 	export IGNORE_SEQUENTIAL_DATA_SETUP=True; \
 	pipenv run python run_in_parallel.py --command_line_args="${TEST_ARGS}"
 
+acceptance_sequential_tests_all: setup
+	export IGNORE_SEQUENTIAL_DATA_SETUP=False; \
+	pipenv run python run_in_sequence.py --command_line_args="${TEST_ARGS}" --test_tags="~donotskipany"
+
```
owencjones: Unless overridden, `TEST_ARGS` will read `--stop`, so this will stop on fail - do you intend that?

If you do, it _might_ be worth taking the opportunity to add a task to not stop?

NickGrantONS: Fail fast, was my thinking

owencjones:
Works for me :+1:
ᐧ

PR #222, opened 2018-11-27
Feature tests for respondent on response status

ryangrundy7:
# Motivation and Context
Feature tests for adding respondent name to the response status page in R-ops.

# What has changed
- Added feature file `display_respondent_alongside_response_status`

# How to test?
- Run with the [Response-ops PR](https://github.com/ONSdigital/response-operations-ui/pull/267) and run the acceptance tests. All should pass
# Links
[Trello](https://trello.com/c/1rcfmtok)

### DIFF
--- a/acceptance_tests/features/display_respondent_alongside_response_status.feature
+++ b/acceptance_tests/features/display_respondent_alongside_response_status.feature
@@ -1,4 +1,5 @@
 @standalone
+@business
 @fixture.setup.data.with.enrolled.respondent.user.and.internal.user.and.new.iac.and.collection.exercise.to.live
 Feature: Internal user can search for a respondent via email
   As an internal user
--- a/acceptance_tests/features/display_respondent_alongside_response_status.feature
+++ b/acceptance_tests/features/display_respondent_alongside_response_status.feature
@@ -1,0 +1,26 @@
+@standalone
+@wip
+Feature: Internal user can search for a respondent via email
+  As an internal user
+  I need to be able to search for a registered respondent via their email
+  So that I can easily find them when dealing with operational calls
+
+#
+#  Background: internal user is already signed in
+#    Given the respondent is signed into their account
+
+  @fixture.setup.data.enrolled.respondent
+  @us211_01
+  Scenario: Respondent field must be present on each Survey displayed on Reporting units for any given Reference
+    Given an external user is enrolled onto a given survey
+    And has not initiated any changes to the "Not started" status
+    When an internal user navigates to Reporting units for that Reference
+    Then no Respondent name should be displayed in the Respondent field
+#
+#  @fixture.setup.data.enrolled.respondent
+#  Scenario: Respondent field must be updated once an external user has completed a survey online
+#    Given an external user is enrolled onto a given survey
+#    And the status is set to Completed
+#    When an internal user navigates to Reporting Units for that reference
+#    Then the Respondent details must be displayed in the Respondent field
+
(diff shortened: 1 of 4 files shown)


acceptance_tests/features/steps/display_respondent_alongside_response_status.py
```diff
@@ -0,0 +1,122 @@
...
+
+@given('the status is set to "Completed by phone"')
+def survey_is_set_to_completed_by_telephone(context):
+    signed_in_respondent(context)
+    go_to_history_tab()
+    assert browser.find_by_text('Completed by phone')
+
+
+@given('the status is set to "No longer required"')
+def survey_set_to_longer_required(context):
```
AndrewTorrance: Probably best not to use quotation marks like this . There is a mechanism in behave tests that can pass these into the tests as a variable. So adding quotation marks and NOT passing a variable will lead to confusion down the line . OR you could add the variable and pass the variable to the browser if appropriate.E.g 

either 

@given('the status is set to No longer required') 

or 

@given('the status is set to "No longer required"')
def survey_set_to_longer_required(context, status):
 signed_in_respondent(context)
    go_to_history_tab()
    assert browser.find_by_text(status)

acceptance_tests/features/steps/display_respondent_alongside_response_status.py
```diff
@@ -0,0 +1,122 @@
...
+def respondent_name_is_not_displayed(_):
+    assert 'Respondent:' in browser.html
+    assert browser.find_by_id('completed-respondent')[0].value is ''
+
+
+@then('the Respondent details must be displayed in the Respondent field')
+def respondent_name_is_displayed(_):
+    assert 'Respondent:' in browser.html
+    respondent = browser.find_by_id('completed-respondent')[0].value
+    assert respondent == 'first_name last_name'
```
AndrewTorrance: Minor point , and not a function of this change ( so no change needed) , but using constant values for first name and last name and then testing that value could give misleading test results under some circumstances

acceptance_tests/features/display_respondent_alongside_response_status.feature
```diff
@@ -0,0 +1,42 @@
+@standalone
+Feature: Internal user can search for a respondent via email
+  As an internal user
+  I need to be able to search for a registered respondent via their email
+  So that I can easily find them when dealing with operational calls
+
+
+  @fixture.setup.data.with.enrolled.respondent.user.and.internal.user.and.new.iac.and.collection.exercise.to.live
```
AdamHawtin: If you have the same tag on all scenarios it can be moved to the feature at the top of the file, it is treated as being on all scenarios in the file.

acceptance_tests/features/display_respondent_alongside_response_status.feature
```diff
@@ -0,0 +1,38 @@
+@standalone
```
AdamHawtin: We're adding `@business` tags to business survey specific features now.

acceptance_tests/features/steps/display_respondent_alongside_response_status.py
```diff
@@ -0,0 +1,115 @@
...
+    mapped_status = map_ce_status('No longer required')
+    change_response_status.update_response_status(mapped_status)
+
+
+@given('a user has partially completed a survey (In progress)')
+def respondent_partially_completes_survey(context):
+    signed_in_respondent(context)
+    browser.click_link_by_id('access_survey_button_1')
+    browser.click_link_by_id('download_survey_button')
+    time.sleep(5)
```
NickGrantONS: I suppose we could do something more elegant than this

ryangrundy7: I'm trying to see if I can get rid of the wait, the only worry is that it attempts to go to the surveys page with the window that closes once it downloads.

AdamHawtin:
Looks like the problem might be that you've use the same scenario title for four of your scenarios, with our method of running them in parallel they need to have unique scenario titles or you can get unexpected behaviour.

PR #223, opened 2018-11-27
Improve acceptance logging

AdeHarrison:
# Motivation and Context
to improve/reduce test log output

# What has changed
Sets a global logging level
Changed some log statement to debug level

# How to test?
make acceptance_tests

# Links
[Trello](https://trello.com/c/v8QqqpMe/377-acceptance-tests-logging)

### DIFF
--- a/run_in_parallel.py
+++ b/run_in_parallel.py
@@ -15,20 +15,29 @@ import time
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from datetime import datetime
 from distutils.util import strtobool
+from logging import getLogger
 from multiprocessing import Queue
-from subprocess import CalledProcessError, PIPE, Popen, TimeoutExpired, check_output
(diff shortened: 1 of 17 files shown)


run_in_parallel.py
```diff
@@ -45,21 +52,30 @@ def parse_arguments():
...
+                        default=DEFAULT_PROCESSES)
     parser.add_argument('--timeout', '-tout', type=int,
                         help='Maximum seconds to execute each scenario. Default = 300', default=300)
 
     args = parser.parse_args()
 
-    # Handle Tag ANDing properly
-    args.tags = create_behave_tags(args.test_tags)
+    # Create behave command line arguments
+    args.behave_args = {
```
AdamHawtin: This is holding both the arguments for the dry run to select scenarios and the actual tests runs, it would be better to split them out, then you can pass the whole object into the relevant function later rather than picking the keys out.

AdeHarrison: fixed

PR #224, opened 2018-11-27
Use thread pool for parallel execution, stop gracefully on test failure

AdamHawtin:
# Motivation and Context
Currently the `run_in_parallel.py` script uses processes to parallelise execution. This PR changes this to use a standard library thread execution pool. It also defaults to gracefully stopping on the first failure, with the option to turn this off. This will cancel all not started scenarios and allow any in progress to complete before exiting.

# What has changed
- Use thread execution pool for parallel test execution
- Stop parallel tests on first failure by default
- Add command line option to not stop on first failure
- Small refactor of parallel _run_scenario function

# How to test?
Run the tests locally, all the parallel tests should still pass and the execution time should not be worse than before the changes.
Introduce a forced test failure and check the tests stop (note they will finish executing any other tests in progress rather than shutting down immediately).

# Links
https://trello.com/c/o1SDodzq/380-acceptance-tests-terminate-parallel-execution-gracefully-on-failure

### DIFF
--- a/run_in_parallel.py
+++ b/run_in_parallel.py
@@ -56,7 +56,7 @@ def parse_arguments():
     parser.add_argument('--test_tags', '-t', help='specify behave tags to run', default=DEFAULT_TAGS)
     parser.add_argument('--timeout', '-tout', type=int,
                         help='Maximum seconds to execute each scenario. Default = 300', default=300)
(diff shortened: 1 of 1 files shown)


AdamHawtin:
[Acceptance Tests: Terminate Parallel execution gracefully on failure](https://trello.com/c/o1SDodzq/380-acceptance-tests-terminate-parallel-execution-gracefully-on-failure)

PR #227, opened 2018-12-03
Alterations for us023

owencjones:
**Don't merge until response-ops branch `reenable-respondent-enrolment` is merged**

# Motivation and Context
US023 added the ability to re-enable enrolment for a repsondent, which meant introducing an extra link - changes test that re-enabling works.

Also added a change to allow screenshots for test diagnostics.

# What has changed
* Altered acceptance tests for US022 to include criterion for US023
* Added `environment.py` code to allow the tests to generate screenshot artifacts
* Set screen size to avoid `splinter` dynamically resizing the virtual window dependent on the contents

# How to test?
* Run the tests to confirm that they run successfully, once the `response-ops` branch has been merged.
* Set the `SCREENSHOT` environment variable to `True`, and force a test to fail.  Confirm that a screenshot is taken of the fail - note that some fails will generate a blank screenshot depending on the kind of fail.

# Links
* https://trello.com/c/StCt8X1m


acceptance_tests/features/environment.py
```diff
@@ -64,6 +66,16 @@
 
 
 def before_all(_):
+    # Clear up old screenshots if in screenshot mode
+    if os.getenv('SCREENSHOT', 'False') == 'True':
+        try:
+            os.unlink(os.path.join(get_screenshot_directory(), '*_failed.png'))
+        except FileNotFoundError:
+            # ignore if there were no previous screenshots to delete
+            print()
```
owencjones: For anyone wondering, this is essentially a `noop` because `flake8` needs an `except` block to contain something, whereas this one is telling it to specifically _ignore_ errors if no files are found.

Note that only `FileNotFoundError` is actually excepted - all other errors will still raise, albeit in a slightly roundabout way.

The generic `except` is functionally identical to not having it, but acts as a verbosity improvement to highlight that the code is raising all but `FileNotFoundError`.

acceptance_tests/features/environment.py
```diff
@@ -115,6 +127,16 @@ def after_scenario(_, scenario):
...
         logger.exception('Failed step', scenario=context.scenario.name, step=step.name)
+        if os.getenv('SCREENSHOT', 'False') == 'True':
+            dir_name = get_screenshot_directory()
+            step_str = step.name
+            file_name = os.path.join(dir_name, step_str + '_failed.png')
+            browser.screenshot(file_name)
+
+
+def get_screenshot_directory():
+    acceptance_tests_directory = os.path.join('..', '..')
```
owencjones: This is probably unnecessary given that we run exclusively in Posix environments, but the use of `os.path.join` proofs us against issues with running on operating systems that don't use `/` as a path separator.

acceptance_tests/features/environment.py
ryangrundy7: In python you can also use a 
```
with suppress(FileNotFoundError):
        code here
```
which i think could work. Not to sure of doing it this way though

ryangrundy7: ^ Ignore that comment won't work while looking for the other exception

acceptance_tests/features/steps/enrolment_status_disable.py
```diff
@@ -4,28 +4,30 @@
 
 
 @given('the internal user is on the ru details page')
+@when(u'the internal user is back on the ru details page')
```
ryangrundy7: Why is this a `u` string?

acceptance_tests/features/environment.py
owencjones: Have changed it to a pass.  Occurred to me that could also just have a check in a generic except block to check if it was a `FileNotFound`

acceptance_tests/features/steps/enrolment_status_disable.py
owencjones: when you miss out an entry an `behave`  then it prompts you to add it, and shows a `u` based string, which in this case isn't needed I guess.

owencjones: Have changed it accordingly

PR #229, opened 2018-12-03
Refactor fixtures and data setup functions

AdamHawtin:
# Motivation and Context
Working with the data setup fixture functions is becoming increasingly difficult, this is an attempt to refactor all that code to make it a little more followable and put all the data setup functions in their most suitable places (importing between `*-utilies.py` packages should now be at a minimum).

# What has changed
- Moved many data setups to different packages
- Removed kwargs from fixture functions, they now only perform one path of data setup rather 
- Refactored all remaining calls to str.format() to f strings
- Removed custom string functions in favour of builtins
- In cases where a respondent doesn't need to be enrolled at all, only set up respondent data (do not enrol then un-enrol)

# How to test?
The tests should execute the same as before, but the fixtures and data setup should be more readable, and data set up utilities should live in appropriate packages.

# Links
https://trello.com/c/hqq9j2gf/389-t389-convert-access-eq-collection-exercise-feature-to-standalone


acceptance_tests/features/fixtures.py
```diff
@@ -128,23 +114,42 @@ def setup_data_with_internal_user_and_collection_exercise_to_created_status(cont
...
 
 
 @fixture
 def setup_data_with_enrolled_respondent_user_and_internal_user_and_new_iac_and_collection_exercise_to_live(context):
     """ Creates default data, an enrolled Respondent, generates a new unused iac
     and waits until collection exercise state = live """
-    setup_data_with_enrolled_respondent_user_and_internal_user(context, generate_new_iac=True,
-                                                               wait_ce_for_state=COLLECTION_EXERCISE_STATUS_LIVE)
-    context.add_cleanup(after_scenario_cleanup, context)
+    create_default_data(context)
```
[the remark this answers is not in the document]
AdamHawtin: Perhaps I was a little overzealous with that, I would still say I prefer it without kwargs deciding the flow of what is setup up. But I can see some places where the same sequence of calls is made, so those could just call the existing fixture.

AdamHawtin:
@AndrewTorrance It is more change than I really wanted to put in at once, the issue I hit was in the utility modules calling each other creating circular imports so the whole package needed a bit of untangling before I could make the changes I wanted to.
I will have a quick pass at factoring out the repeated sequences in fixtures to make it DRY.

acceptance_tests/features/environment.py
```diff
@@ -138,3 +137,9 @@ def get_survey_type(tags):
         return "Social"
 
     return 'Business'
+
+
+def setup_sequential_data_for_test():
```
AdeHarrison: not sure this belongs here?

acceptance_tests/features/fixtures.py
```diff
@@ -1,20 +1,20 @@
...
-                                    create_test_business_collection_exercise, \
-                                    create_test_survey, \
-                                    create_unenrolled_respondent
+from acceptance_tests.features.pages import sign_out_internal
+from common.collection_exercise_utilities import create_business_survey_period, create_social_survey_period, \
+    generate_collection_exercise_dates_from_period, generate_new_enrolment_code_from_existing_code, \
+    make_user_description
+from common.internal_user_utilities import create_internal_user_login_account
+from common.respondent_utilities import create_enrolled_respondent_for_the_test_survey, create_respondent, \
+    create_respondent_data, create_respondent_email_address, create_respondent_user_login_account, create_ru_reference, \
```
AdeHarrison: will fail checkstyle as 1 char too long

acceptance_tests/features/environment.py
AdamHawtin: I'm not sure where it belongs, we don't have any specific place for sequential test data setup so putting it with the hook where it's called seemed to make sense?

AdeHarrison: doesn't make a lot of difference as it will eventually be deleted but thought maybe fixures.py more appropriate as all setup happens there?

common/respondent_utilities.py
```diff
@@ -80,3 +91,36 @@ def register_respondent(survey_id, period, username, ru_ref=None, verified=True)
...
+    create_respondent(user_name=context.respondent_email, enrolment_code=context.iac, phone_number=context.phone_number)
+    create_respondent_user_login_account(context.respondent_email)
+
+
+def create_unenrolled_respondent(context):
+    create_respondent_data(context)
+    create_respondent(user_name=context.respondent_email, enrolment_code=context.iac, phone_number=context.phone_number)
+    create_respondent_user_login_account(context.respondent_email)
+
+    unenrol_respondent_in_survey(context.survey_id)
```
AdeHarrison: still unenrolling after, is this intentional?

AdamHawtin: Yes, this PR already got bigger than I was intended and it does work as it is.

PR #238, opened 2018-12-13
Added test that respondent now replies to a specific internal user

AndrewTorrance:
# Motivation and Context
New functionality added in Fronstage to set the msg_to in secure message to a specific internal person when replying to a message from an internal person. So that the internal user can see the name of a person associated with the conversation

# What has changed
New test when a respondent replies to a message from the internal user checking that the msg_to has the expected internal user id

# How to test?

run the new scenario
  Scenario: When a respondent replies to a message it is sent to the same internal user that sent the original message

# Links
 https://trello.com/c/PGjxHXYA

### DIFF
--- a/acceptance_tests/features/steps/inbox_internal.py
+++ b/acceptance_tests/features/steps/inbox_internal.py
@@ -128,6 +128,6 @@ def closed_tab_is_visible(_):
 
 
 @then('The To field should be the name of the internal user who sent the message')
-def to_filed_holds_name_of_internal_user_who_sent_the_message(context):
+def to_field_holds_name_of_internal_user_who_sent_the_message(context):
     first_message = inbox_internal.get_unread_messages()[0]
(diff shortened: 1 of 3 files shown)


acceptance_tests/features/steps/inbox_internal.py
```diff
@@ -124,3 +125,9 @@ def pagination_links_available(_):
 @then('they can see the closed tab')
 def closed_tab_is_visible(_):
     assert inbox_internal.closed_tab_present()
+
+
+@then('The To field should be the name of the internal user who sent the message')
+def to_filed_holds_name_of_internal_user_who_sent_the_message(context):
```
jcox-dev:
```suggestion
def to_field_holds_name_of_internal_user_who_sent_the_message(context):
```

AndrewTorrance: Doh !

PR #253, opened 2018-12-28
Added tests for marking messages unread

AndrewTorrance:
# Motivation and Context
The requirement was that an internal user can mark a message as unread. So that they can use the read status as a proxy indicator that a message needs to be followed up. 

Some complications arise because now we have the ability to send messages from a respondent to a specific internal user.

# What has changed
The use cases need some explaining. We have a mark as unread link available on the last message in a conversation from a respondent. This only shows if the message was sent to GROUP or to the specific internal user who is looking at the message. IF that link shows then the internal user can select it to mark the message as unread and return to the list of conversations.  

There is also a pair of back links on the page. In earlier versions pressing these would return the user to the conversation view and mark the message as read. Now this has slightly different functionality. If the message was sent to group or to the internal user who is logged in then pressing back will mark the message as read. If however the message went to specific internal user and that is NOT the logged in user then pressing back does not mark the message as read. 

# How to test?
Run the acceptance tests .

# Links
https://trello.com/c/Bvakzk6d


acceptance_tests/features/view_full_thread_internal.feature
```diff
@@ -33,3 +33,62 @@ Feature: View conversation thread
...
+     And  the expected message is displayed
+     And  they are able to distinguish that the message is unread
+
+  Scenario: Message sent from internal, respondent replies , user who sent original message can mark as unread
+    Given the internal user has received a message
+      And  an internal user responds and respondent signs in
+      And the respondent navigates to their inbox
+      And the respondent replies to first conversation
+    When The internal user is already signed in
+      And the internal person views the message
```
ryangrundy7: I think changing this to internal user instead of internal person sounds better

PR #256, opened 2019-01-22
Return to page when unread

AndrewTorrance:
# Motivation and Context
Pressing mark unread was returning user to wrong page. Also some tests worked locally and not remotely 

# What has changed
Added extra tests . 

# How to test?
Run tests

### DIFF
--- a/acceptance_tests/features/pages/create_message_internal.py
+++ b/acceptance_tests/features/pages/create_message_internal.py
@@ -1,6 +1,6 @@
 from acceptance_tests import browser
 from acceptance_tests.features.pages.reporting_unit import click_data_panel
-from common.browser_utilities import element_by_id_on_page, wait_for
+from common.browser_utilities import wait_for_element_by_id
 
 from config import Config
 
@@ -46,7 +46,7 @@ def click_cancel_button():
 
 
 def click_back_link():
-    wait_for(element_by_id_on_page, 5, 1, 'back-link')
+    wait_for_element_by_id(element_id='back-link', timeout=5, retry=1)
     browser.find_by_id('back-link').click()
 
 
@@ -79,7 +79,7 @@ def clear_message_body_text():
 
 
 def get_first_flashed_message():
-    wait_for(element_by_id_on_page, 5, 1, "flashed-message-1")
+    wait_for_element_by_id(element_id="flashed-message-1", timeout=5, retry=1)
     return browser.find_by_id("flashed-message-1").value
 
 
--- a/acceptance_tests/features/pages/create_message_internal.py
+++ b/acceptance_tests/features/pages/create_message_internal.py
@@ -79,6 +79,7 @@ def clear_message_body_text():
 
 
 def get_first_flashed_message():
+    wait_for(element_by_id_on_page, 5, 1, "flashed-message-1")
     return browser.find_by_id("flashed-message-1").value
 
 
--- a/acceptance_tests/features/pages/create_message_internal.py
+++ b/acceptance_tests/features/pages/create_message_internal.py
@@ -1,5 +1,7 @@
 from acceptance_tests import browser
 from acceptance_tests.features.pages.reporting_unit import click_data_panel
+from common.browser_utilities import element_by_id_on_page, wait_for
+
 from config import Config
 
 
@@ -44,6 +46,7 @@ def click_cancel_button():
 
 
 def click_back_link():
+    wait_for(element_by_id_on_page, 5, 1, 'back-link')
     browser.find_by_id('back-link').click()
 
 
--- a/acceptance_tests/features/pages/internal_conversation_view.py
+++ b/acceptance_tests/features/pages/internal_conversation_view.py
@@ -1,6 +1,6 @@
 from acceptance_tests import browser
 from config import Config
-from common.browser_utilities import element_by_id_on_page, wait_for
+from common.browser_utilities import wait_for_element_by_id
 
 
 def go_to():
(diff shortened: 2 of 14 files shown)


common/browser_utilities.py
```diff
@@ -12,15 +12,23 @@ def is_text_present_with_retry(text, retries=3, delay=1):
     return False
 
 
-def wait_for(fn, timeout_secs, retry_secs=3.0):
+def wait_for(fn, timeout_secs, retry_secs=3.0, argv=[]):
```
insacuri: argv could and should have better name (as with most things in this function).

Explain how and when argv should be used in the docstring.

common/browser_utilities.py
```diff
@@ -12,15 +12,23 @@ def is_text_present_with_retry(text, retries=3, delay=1):
...
 
     while not ret_val and datetime.utcnow() < timeout:
-        ret_val = fn()
+        ret_val = fn(argv) if argv else fn()
         if not ret_val:
             time.sleep(retry_secs)
     return ret_val
+
+
+def named_element_on_page(name):
```
insacuri: Doc string explaining what it's for and what the variables are used for

common/browser_utilities.py
```diff
@@ -12,15 +12,23 @@ def is_text_present_with_retry(text, retries=3, delay=1):
...
+    return True if browser.find_by_name(name) else False
+
+
+def element_by_id_on_page(element_id):
```
insacuri: Doc string explaining what it's for and what the variables are used for

common/browser_utilities.py
insacuri: Also, argv is an empty list here, but above it's passed a string.  The default for this should probably be `None` as the default usually gives a hint as to what a variable should be and this is just confusing.

acceptance_tests/features/steps/reply_to_message_internal.py
```diff
@@ -140,17 +140,21 @@ def url_is_for_my_messages(_):
...
+    internal_user_taken_to_specific_page(context, page)
 
     go_to_thread()
 
 
+@given("the internal user goes to page '{page}'")
+def internal_user_taken_to_specific_page(context, page):
+    inbox_internal.go_to_using_context(context)
+    for page_num in range(1, int(page)):
+        wait_for(named_element_on_page, 5, 1, 'next')
```
insacuri: Whilst you're using this function, I strongly feel like we should be using it in a way that's easier to understand (and hopefully this version proliferate). 
This SHOULD read something like 
`wait_for(named_element_on_page, timeout=5, limit=1, argv='next')`.  It's really really unclear what those 2 numbers mean without going directly to the source so using kwargs here will be helpful.

acceptance_tests/features/steps/inbox_internal.py
```diff
@@ -117,18 +116,10 @@ def internal_user_has_unread_message_in_inbox(context):
 
 @then('they are able to distinguish that the message is unread')
 def internal_user_can_distinguish_the_message_is_unread(_):
-    wait_for(_validate_message_unread_on_page, 5, 1)
+    wait_for(named_element_on_page, 8, 1, 'message-unread')
```
insacuri: This should read `wait_for(named_element_on_page, timeout=5, limit=1, argv='message-unread')`

acceptance_tests/features/pages/internal_conversation_view.py
```diff
@@ -1,12 +1,14 @@
...
 from config import Config
+from common.browser_utilities import element_by_id_on_page, wait_for
 
 
 def go_to():
     browser.visit(f"{Config.RESPONSE_OPERATIONS_UI}/messages")
 
 
 def go_to_thread():
+    wait_for(element_by_id_on_page, 8, 1, 'message-link-1')
```
insacuri: This should be changed to use kwargs (see other comments for examples)

acceptance_tests/features/pages/view_and_reply_conversation_external.py
```diff
@@ -20,6 +21,7 @@ def get_sender_from_conversation_message(conversation_message):
 
 def go_to_first_conversation_in_message_box():
     browser.visit(f"{Config.FRONTSTAGE_SERVICE}/secure-message/threads")
+    wait_for(element_by_id_on_page, 5, 1, "open-conversation-link-1")
```
insacuri: This should be changed to use kwargs (see other comments for examples)

acceptance_tests/features/pages/create_message_internal.py
```diff
@@ -44,6 +46,7 @@ def click_cancel_button():
 
 
 def click_back_link():
+    wait_for(element_by_id_on_page, 5, 1, 'back-link')
```
insacuri: This should be changed to use kwargs (see other comments for examples)

acceptance_tests/features/pages/create_message_internal.py
```diff
@@ -76,6 +79,7 @@ def clear_message_body_text():
 
 
 def get_first_flashed_message():
+    wait_for(element_by_id_on_page, 5, 1, "flashed-message-1")
```
insacuri: This should be changed to use kwargs (see other comments for examples)

common/browser_utilities.py
AndrewTorrance: Agreed , argv use is poor. tidying it up to be more explicit

common/browser_utilities.py
AndrewTorrance: Done

acceptance_tests/features/steps/reply_to_message_internal.py
AndrewTorrance: I know what your saying but since the function MAY take a variable number of arguments I cant pass named variables and variable length args in same call in a neat way ? So what I have done is create two helper functions that can take named args which covers us most of the time

AndrewTorrance:
> It all passed for me locally 👍 If it doesn't make the tests pass on the pipeline, might be worth `WebDriverWait` just to see if that makes it pass instead.

Web driver wait doesn't do a retry . There was a bug in wait which would give a false positive. I think I have caught that . I guess we will see.....

PR #257, opened 2019-01-30
Skipping already added survey test

ryangrundy7:
# Motivation and Context
We need to revert a change on ras-frontstage and leaving this test in will break the pipeline. For the time being, we're gonna skip the test.
# What has changed
- Added skip tag to `User is trying to add a survey using a new iac code that they have previously added`

# How to test?
- Run the acceptance tests with [the revert branch](https://github.com/ONSdigital/ras-frontstage/tree/revert-already-added-survey) and make sure all tests pass

### DIFF
--- a/acceptance_tests/features/add_a_survey.feature
+++ b/acceptance_tests/features/add_a_survey.feature
@@ -10,11 +10,11 @@ Feature: Add a survey
 
 
   @fixture.setup.data.with.enrolled.respondent.and.additional.iac
+  @skip
   Scenario: User is trying to add a survey using a new iac code that they have previously added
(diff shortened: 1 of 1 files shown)


acceptance_tests/features/add_a_survey.feature
```diff
@@ -10,12 +10,12 @@ Feature: Add a survey
 
 
   @fixture.setup.data.with.enrolled.respondent.and.additional.iac
+  @skip
```
owencjones: Have spoken to @ryangrundy7 and confirmed that this test is left in place to avoid issues with VCS down the line.

Added tech debt card to address this later: https://trello.com/c/R7M5uhMM

PR #259, opened 2019-02-01
723 add asserts to wait

AndrewTorrance:
# Motivation and Context
Tests passing locally are failing in CI. With no indication why. We suspect that there may be timing issues. This will add an assert to each wait such that if after the wait period the element is still not present then an assert will be raised and visible in the output

# What has changed
Added assert to waits
Added wait for by class as the test that was using it passed locally and failed in CI , and used to use a by_class method

# How to test?
Run tests

### DIFF
--- a/common/browser_utilities.py
+++ b/common/browser_utilities.py
@@ -13,7 +13,7 @@ def is_text_present_with_retry(text, retries=3, delay=1):
 
 
 def wait_for_element_by_name(name, timeout, retry):
-    """Waits for the named element to appear on the page
+    """Waits for the named element to appear on the page, asserts if not present after timeout
 
     Parameters:
         name (string): The name of the element to wait for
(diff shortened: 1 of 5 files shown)


common/browser_utilities.py
```diff
@@ -38,7 +62,10 @@ def wait_for_element_by_id(element_id, timeout, retry):
...
         boolean: True if element found else False
     """
-    return wait_for(_element_by_id_on_page, timeout, retry, element_id)
+    ret_val = wait_for(_element_by_id_on_page, timeout, retry, element_id)
+    assert ret_val
+
+    return ret_val
 
 
 def wait_for(fn, timeout, retry_after, *argv):
```
owencjones: "No" is a perfectly valid repsonse, but do you think we could do with having some defaults for this function?

common/browser_utilities.py
```diff
@@ -64,6 +91,7 @@ def wait_for(fn, timeout, retry_after, *argv):
         ret_val = fn(*argv) if argv else fn()
         if not ret_val:
             time.sleep(retry_after)
+
     return ret_val
```
owencjones: Might have gone for a retry count, rather than a timeout, but both work.

PR #269, opened 2019-03-25
Fixing tests for clocks going forward

ryangrundy7:
# Motivation and Context
The tests for checking the collection exercise event dates puts the return_by and exercise_end date 10 days in the future which then gets changed to the BST time. This change should fix it so it saves it as UTC for the database and display the current timezone on the page.
# What has changed
- Tests should work with timezone differences of UTC and BST.

# How to test?
- Run the acceptance tests
- Run `view_collection_exercise_details.feature` with the date in april and the test should still pass.

### DIFF
--- a/acceptance_tests/features/fixtures.py
+++ b/acceptance_tests/features/fixtures.py
@@ -1,9 +1,10 @@
 from behave import fixture
+from dateutil import tz
 
 from acceptance_tests.features.pages import sign_out_internal
 from common.collection_exercise_utilities import create_business_survey_period, create_social_survey_period, \
     execute_collection_exercises, generate_collection_exercise_dates_from_period, \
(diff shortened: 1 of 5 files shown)


acceptance_tests/features/fixtures.py
```diff
@@ -162,6 +162,10 @@ def _setup_data_with_internal_user_and_collection_exercise_to_specific_status(co
     response = create_test_business_collection_exercise(survey_id, period, short_name, ce_name,
                                                         survey_type, stop_at_state=stop_at_state)
 
+    for key, date in response['dates'].items():
```
mfcrocker: Rather than doing this, wouldn't it be better to set the TZ info in `generate_collection_exercise_dates_from_period`?

mfcrocker: Is this still needed after the changes @ryangrundy7

PR #271, opened 2019-04-03
Fixes for broken acceptance tests

ryangrundy7:
# Motivation and Context
The pipeline has been broken for a couple of weeks and when testing it in the pipeline, there's some tests failing because of recent changes gone into response-operations-ui. These changes should hopefully get the pipeline green again.
# What has changed
- Changed test to find new alert message.
- Use local timezone for collection exercise events replace .
- Changed `re_send_verification.feature` to be a standalone test because it creates its own data.

# How to test?
- Run the acceptance tests and all should pass

### DIFF
--- a/common/collection_exercise_utilities.py
+++ b/common/collection_exercise_utilities.py
@@ -3,6 +3,7 @@ from datetime import datetime, timedelta
 from logging import getLogger
 
 from dateutil import tz
+from dateutil.tz import tzlocal
 from structlog import wrap_logger
 
 import common.string_utilities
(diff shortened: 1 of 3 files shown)


common/collection_exercise_utilities.py
```diff
@@ -100,7 +101,7 @@ def generate_collection_exercise_dates_from_period(period):
     """Generates a collection exercise events base date from the period supplied."""
 
     now = datetime.now()
-    now = now.replace(tzinfo=tz.gettz('Europe/London')).astimezone(tz.gettz('UTC'))
+    now = now.replace(tzinfo=tzlocal()).astimezone(tz.gettz('UTC'))
```
mfcrocker: Will this cause problems based on which server location we host at?

ryangrundy7: For the time being this way should work, gonna put a card on the board to look at the way we do timezone conversion in the acceptance tests

PR #275, opened 2019-06-20
937 additional waits and checks

AndrewTorrance:
# Motivation and Context
Acceptance tests were failing seemingly randomly. This was an investigation to try and find why.

# What has changed
What was found was : 
1) Chrome driver in headless drops the session after 30 seconds , firefox does not
2)Several tests had finger in the air sleep statements , several of these changed to await for new elements , url changes etc
3) The remaining common failure is a 409 from the survey service which should be fixed by an upcoming release of survey service and tests

This release uses an environment variable called WEBDRIVER . This defaults to chrome , but can (and should) be over ridden to firefox. The dev can switch between firefox and chrome as and when they wish. We can then loop back around after everyone is using firefox and set the default to firefox.

Note gecko writes to geckodriver.log so you may wish to clear that down periodically

# How to test?

1) Install Firefox 
2) Install gecko driver ( e.g brew install geckodriver ) , make sure its on path.
3) export WEBDRIVER=firefox (in .bashrc ?) 

Run acceptance tests . You can switch back to chrome at any time by either deleting WEBDRIVER environment variable or setting it to 'chrome' . 
NB. Remember to set WEBDRIVER in env variables in whatever IDE you use  else you will carry on with chrome

Errors of the type : HTTPError: 409 Client Error: Conflict for url: http://localhost:8080/surveys are expected and should be cleaned up by a later change. But aside from that the tests run a lot more reliably

# Links
https://trello.com/c/QSoSFAU2

### DIFF
--- a/acceptance_tests/__init__.py
+++ b/acceptance_tests/__init__.py
@@ -20,9 +20,9 @@ def web_driver_connection_error(e):
 
 @retry(retry_on_exception=web_driver_connection_error, wait_fixed=1000, stop_max_attempt_number=30)
 def create_browser():
-    driver_type = os.getenv('WEBDRIVER', 'chrome')
+    driver_type = os.getenv('WEBDRIVER', 'chrome').lower()
     headless = os.getenv('HEADLESS', 'True') == 'True'
-    if driver_type.lower() == 'firefox':
+    if driver_type == 'firefox':
         return Browser('firefox', headless=headless)
     else:
         if headless:
--- a/acceptance_tests/__init__.py
+++ b/acceptance_tests/__init__.py
@@ -22,7 +22,6 @@ def web_driver_connection_error(e):
 def create_browser():
     driver_type = os.getenv('WEBDRIVER', 'firefox')
     headless = os.getenv('HEADLESS', 'True') == 'True'
-
     if driver_type.lower() == 'firefox':
         return Browser('firefox', headless=headless)
     else:
--- a/acceptance_tests/__init__.py
+++ b/acceptance_tests/__init__.py
@@ -20,14 +20,19 @@ def web_driver_connection_error(e):
 
 @retry(retry_on_exception=web_driver_connection_error, wait_fixed=1000, stop_max_attempt_number=30)
 def create_browser():
+    driver_type = os.getenv('WEBDRIVER', 'firefox')
+    headless = os.getenv('HEADLESS', 'True') == 'True'
 
-    if os.getenv('HEADLESS', 'True') == 'True':
-        chromedriver_binary.add_chromedriver_to_path()
-        chrome_options = webdriver.ChromeOptions()
-        chrome_options.add_argument("--no-sandbox")
-        return Browser('chrome', headless=True, options=chrome_options)
+    if driver_type.lower() == 'firefox':
+        return Browser('firefox', headless=headless)
     else:
-        return Browser('chrome')
+        if headless:
+            chromedriver_binary.add_chromedriver_to_path()
+            chrome_options = webdriver.ChromeOptions()
+            chrome_options.add_argument("--no-sandbox")
+            return Browser('chrome', headless=True, options=chrome_options)
+        else:
+            return Browser('chrome')
 
 
 browser = create_browser()
--- a/acceptance_tests/__init__.py
+++ b/acceptance_tests/__init__.py
@@ -20,7 +20,7 @@ def web_driver_connection_error(e):
 
 @retry(retry_on_exception=web_driver_connection_error, wait_fixed=1000, stop_max_attempt_number=30)
 def create_browser():
-    driver_type = os.getenv('WEBDRIVER', 'firefox')
+    driver_type = os.getenv('WEBDRIVER', 'chrome')
     headless = os.getenv('HEADLESS', 'True') == 'True'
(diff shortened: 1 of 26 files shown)


acceptance_tests/__init__.py
```diff
@@ -20,14 +20,18 @@ def web_driver_connection_error(e):
 
 @retry(retry_on_exception=web_driver_connection_error, wait_fixed=1000, stop_max_attempt_number=30)
 def create_browser():
-
-    if os.getenv('HEADLESS', 'True') == 'True':
-        chromedriver_binary.add_chromedriver_to_path()
-        chrome_options = webdriver.ChromeOptions()
-        chrome_options.add_argument("--no-sandbox")
-        return Browser('chrome', headless=True, options=chrome_options)
+    driver_type = os.getenv('WEBDRIVER', 'chrome')
```
mfcrocker: If we do `os.getenv('WEBDRIVER', 'chrome').lower()` here we can just pass driver_type straight in below rather than passing in chrome or firefox.

acceptance_tests/__init__.py
```diff
@@ -20,14 +20,18 @@ def web_driver_connection_error(e):
...
+    headless = os.getenv('HEADLESS', 'True') == 'True'
+    if driver_type.lower() == 'firefox':
+        return Browser('firefox', headless=headless)
     else:
-        return Browser('chrome')
+        if headless:
+            chromedriver_binary.add_chromedriver_to_path()
+            chrome_options = webdriver.ChromeOptions()
+            chrome_options.add_argument("--no-sandbox")
+            return Browser('chrome', headless=True, options=chrome_options)
```
mfcrocker: Not sure why the firefox one passes `headless=headless` but this passes `headless=True`; headless will always be True here

acceptance_tests/features/pages/collection_exercise.py
```diff
@@ -19,15 +22,22 @@ def _is_state(first_state, second_state):
 
 
 def go_to(survey):
-    browser.visit(f'{Config.RESPONSE_OPERATIONS_UI}/surveys/{survey}')
+
```
mfcrocker: :art:

acceptance_tests/features/steps/sign_out_internal.py
```diff
@@ -1,15 +1,17 @@
...
 
 @then('the user is logged out and shown the sign in page and they see a successfully signed out message')
 def view_home_page(_):
-    sign_out_internal.signed_out_successfully_message()
+    signed_out_successfully_message()
 
 
 @given('they click the sign out link')
 @when('they click the sign out link')
+@given('the internal user signs out')
```
mfcrocker: Does it make sense for this to have 5 different contexts over all 3 of given/when/then? This isn't a rhetorical question, I'm just surprised that we've managed to shoehorn in this action for all of them. At the very least, it probably makes sense to standardise whether we're calling it `they click the sign out link` or `the internal user signs out`

acceptance_tests/__init__.py
AndrewTorrance: The drivers seem to have different config options . Changing to driver_type = os.getenv('WEBDRIVER', 'chrome').lower() so we know its state on entry but I think we need to retain different config code

acceptance_tests/__init__.py
AndrewTorrance: Its the different ways chrome and firefox are configured

acceptance_tests/features/steps/sign_out_internal.py
AndrewTorrance: Fair catch , knocking it down to 2 , but have to rerun tests

acceptance_tests/__init__.py
mfcrocker: That shouldn't stop us passing in `headless=headless` rather than `headless=True` if the variable is guaranteed to be `True`

acceptance_tests/__init__.py
mfcrocker: It makes no sense to pass in `chrome` or `firefox` as string literals when we can pass in the variable `driver_type`

acceptance_tests/features/enrolment-status-disable-enable.feature
```diff
@@ -28,16 +28,22 @@ Feature: Disable respondent enrolment status
...
     Then the respondent should not be able to view the disabled enrolment
 
   @us022_s03
   Scenario: Internal user disables a respondents enrolment the other respondent should still see survey
     Given the respondent with email "disable_respondent_3@email.com" is enrolled and active
-    And the respondent with email "disable_respondent_4@email.com" is enrolled and active
-    And the internal user disables enrolment for respondent with email "disable_respondent_3@email.com"
-    When the respondent with email "disable_respondent_4@email.com" views their survey todo list
+      And the respondent with email "disable_respondent_4@email.com" is enrolled and active
+      And the internal user signs out
```
rhiGriff: Why are we signing out the internal user before signing them back in again in the step below?

AndrewTorrance: Removed it. It came about during finding debugging the chrome headless issue. By logging out and in it reset the time before it would drop the session and hence the test would pass

PR #279, opened 2019-07-11
Fixes for new survey changes

ChristopherMShelton:
# Motivation and Context
acceptance test changes for survey service

# What has changed
changed tests to acomodate changes

# How to test?
run ATs

# Links
https://trello.com/c/GvoKYQUd/1077-merge-in-ryans-acceptance-test-changes-after-survey-service-changes-have-made-it-to-master

### DIFF
--- a/acceptance_tests/features/social_sign_in_internal.feature
+++ b/acceptance_tests/features/social_sign_in_internal.feature
@@ -25,17 +25,17 @@ Feature: Internal user signs in to social UI
     Given the user has an active account and is assigned a username and password (social)
     When they enter an incorrect username and password (social)
     Then the user is notified that an authentication error has occurred (social)
-
+  @skip
   Scenario: User attempts sign in and is notified that they are required to enter a password (social)
     Given the user has an active account and is assigned a username and password (social)
     When they enter a correct username and no password (social)
     Then the user is notified that a password is required (social)
-
+  @skip
   Scenario: User attempts sign in and is notified that they are required to enter a username (social)
     Given the user has an active account and is assigned a username and password (social)
     When they enter no username and a correct password (social)
     Then the user is notified that a username is required (social)
-
+  @skip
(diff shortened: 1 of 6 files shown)


acceptance_tests/features/social_sign_in_internal.feature
```diff
@@ -25,17 +25,17 @@ Feature: Internal user signs in to social UI
     Given the user has an active account and is assigned a username and password (social)
     When they enter an incorrect username and password (social)
     Then the user is notified that an authentication error has occurred (social)
-
+  @skip
```
rhiGriff: Can we put in a comment as to why these are being skipped please?

mfcrocker: 👍

controllers/collection_exercise_controller.py
```diff
@@ -269,10 +269,11 @@ def create_business_collection_exercise_to_ready_for_review_state(survey_id, per
 
     if eq_ci:
         ci_controller.load_and_link_eq_collection_instrument(survey_id, collection_exercise_id,
-                                                             form_type=ru_ref, eq_id=ru_ref)
+                                                             form_type='0001', eq_id=ru_ref)
```
necrophonic: Is there a reason these are hardcoded to `0001`? Could be worth a comment to avoid accidental changing and bugs in the future

acceptance_tests/features/social_sign_in_internal.feature
ChristopherMShelton: added a lil' doc string

ChristopherMShelton:
> Happy with this, though would personally like to see a reply to @necrophonic

spoke to him about it, hardcoding the 0001 because its a form type not an RU ref. Also most of the codebase uses 0001 type for forms. Only real alternative would be to random a formtype form an array of them but there's really no benefit to it.

PR #288, opened 2019-10-21
validating correct tab when back via flash message

AndrewTorrance:
# Motivation and Context
response ops was not tracking the tab of origin of a conversation and was thus returning to open when a different tab was more appropriate. Also it was returning to the wrong tab. Most of the work was tested within response ops , but one scenario using the link in the flash message could not be easily tested that way so a new acceptance test was created

# What has changed
Deleted acceptance tests that replicated tests added to the response_ops repo.
Changed the tab definition to be a multi value conversation tab variable instead of bools to reflect changes in response ops. Refactored some tab selections to use a single call and be more DRY.

# How to test?
Run acceptance tests 

# Links
https://trello.com/c/mego4FuA
https://github.com/ONSdigital/response-operations-ui/compare/secure-message-return-to-tab?expand=1

### DIFF
--- a/acceptance_tests/features/pages/inbox_internal.py
+++ b/acceptance_tests/features/pages/inbox_internal.py
@@ -3,35 +3,18 @@ from common.browser_utilities import wait_for_url_matches
 from config import Config
 
 
-def go_to(context):
-    browser.visit(f"{Config.RESPONSE_OPERATIONS_UI}/messages/{context.short_name}")
+# def go_to(context):
+#     browser.visit(f"{Config.RESPONSE_OPERATIONS_UI}/messages/{context.short_name}")
 
 
-def go_to_closed():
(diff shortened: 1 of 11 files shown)


acceptance_tests/features/pages/inbox_internal.py
```diff
@@ -3,35 +3,18 @@
 from config import Config
 
 
-def go_to(context):
-    browser.visit(f"{Config.RESPONSE_OPERATIONS_UI}/messages/{context.short_name}")
-
-
-def go_to_closed():
-    browser.visit(f"{Config.RESPONSE_OPERATIONS_UI}/messages/Bricks?is_closed=true")
+# def go_to(context):
```
mfcrocker: If this PR works fine, can we just delete these 2 methods now?

PR #295, opened 2020-02-18
Explicit waits

jdfox:
PR
# Motivation and Context
From ticket “Fix acceptance tests”, change was needed to fix timeouts during acceptance testing to prevent continual failure of tests on unchanged code and data.

# What has changed
Added timeouts to Selenium tests.
No refactoring was required.
No new tests written, as this fix is on a set of tests. 
# How to test?
**For all tests**:
make test 

**For just create_collection_exercise**:
make setup
make TEST_TARGET=acceptance_tests/features/create_collection_exercise.feature run_tests

# Links
https://trello.com/c/9FIFx6cf/1631-fix-acceptance-tests

### DIFF
--- a/acceptance_tests/features/pages/collection_exercise.py
+++ b/acceptance_tests/features/pages/collection_exercise.py
@@ -3,6 +3,10 @@ from functools import partial
 from acceptance_tests import browser
 from config import Config
 
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support.ui import WebDriverWait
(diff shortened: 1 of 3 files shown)


jdfox:
Note that the test "Create collection exercise" currently still times out on "click_collection_exercise_created_banner". I temporarily bumped the timeout value up to 240 seconds and it still failed, so I suspect that the failed test is either pointing to an actual problem in the code, or is testing something that will time out forever. Could reviewers with strong knowledge of ras_party please comment?

